PR #15, opened 2021-09-10
Background

Soham041201:
Changed the background color to sky blue :)


src/index.css
```diff
@@ -5,9 +5,13 @@ body {
...
   -webkit-font-smoothing: antialiased;
   -moz-osx-font-smoothing: grayscale;
+  background-color:skyblue;
 }
 
 code {
   font-family: source-code-pro, Menlo, Monaco, Consolas, 'Courier New',
     monospace;
 }
+.image{
```
eddiejaoude, line 15: This has come in by PrimeReact so is no longer required

PR #16, opened 2021-09-10
Created separate component for title to use at multiple places and ea…

veerbal1:
Created separate component for title to use at multiple places and easily modification for future updates


eddiejaoude (approved):
Thank you 👍

PR #120, opened 2021-09-13
Avatar url Updated to .png Url 

rohan-kulkarni-25:


FayasNoushad:
Good work 👍

public/data/Kaushik.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "Kaushik",
   "bio": "Student",
-  "avatar": "https://avatars.githubusercontent.com/KaushikKC",
+  "avatar": "https://github.com/KaushikKC.png",
```
nhcarrigan, line 4:
```suggestion
  "avatar": "https://github.com/KaushikK.png",
```

Discrepancy between GitHub username and avatar username?

nhcarrigan (changes requested):

public/data/Kaushik.json
rohan-kulkarni-25: `https://github.com/KaushikKC.png` This one url is correct as this person opened issue Both url are available but the one with kkc is right person

nhcarrigan: I think it's confusing to have the GitHub URL different from the avatar URL - maybe we should leave this one alone and allow them to update it themselves.

rohan-kulkarni-25: > I think it's confusing to have the GitHub URL different from the avatar URL - maybe we should leave this one alone and allow them to update it themselves.

Yeah we can do that. I will commit their own profile url which they uploaded and ask them to update it

eddiejaoude (approved):
Thank you 👍

PR #174, opened 2021-09-14
Added Hover Effect on Home Page #144

### LINKED ISSUE #144: Add Responsive Card type hover effect in the home where all member's profile are showing...
So,
     1. When we hover in the main home page where all member's profile are showing...
     2. The pointed profile will hover with responsive card hover effect with the profile photo and the username.

aayush89890:
Fixes #144


eddiejaoude:
Thank you! But this features has already been added

PR #254, opened 2021-09-20
feat: milestones (#252)

### LINKED ISSUE #252: Add Milestone Feature
Milestones for users, related to the existing Polywork-Feature.

The user should be able to add milestones to their profile. The milestones could be displayed in the Timeline-Component from PrimeReact.

eddiejaoude:
Fixes #252 

Screenshot 2021-09-20 at 02 22 07

<a href="https://gitpod.io/#https://github.com/EddieHubCommunity/LinkFree/pull/254"></a>


schmelto:
Mobile view is looking not so good.

voidz7:
video

Looks good to me ⚡ . We may need to add more docs about adding ur data for milestones ....

Cahllagerfeld:
I feel like we should sort the milestones descending to the date? What do you think about that?😅

voidz7:
> I feel like we should sort the milestones descending to the date (So the newest is on top, and the oldest on the bottom)? What do you think about that?sweat_smile

ig the person who is making their own milestone has the ability to format it descending or ascending  . Right ?

Cahllagerfeld:
In my opinion we should write a function which sorts the milestones. Imagine the following scenario:
Someone who initally added 5 milestons wants to add another one. When sorting them manually, he has to maybe restructure all milestones so the one that should be inserted fits chronologically.

When adding a function everyone can just add one as the last entry and everything gets done automatically afterwards.
But thats just my opinion, I'm keen on hearing other opinions :D

CodingSpecies:
I am not sure, but couldn't it be stored in like a data type, array, or object, and then it auto generates the format, into the milestone?

voidz7:
> In my opinion we should write a function which sorts the milestones. Imagine the following scenario:
> Someone who initally added 5 milestons wants to add another one. When sorting them manually, they have to maybe restructure all milestones so the one that should be inserted fits chronologically.
> 
> When adding a function everyone can just add one as the last entry and everything gets done automatically afterwards.
> But thats just my opinion, I'm keen on hearing other opinions :D

oo this would be cool then rather worrying about the order u can keep adding your milestones in the json with date and we just have a function which keeps sorting them ! vouch for this .

eddiejaoude:
Great feedback everyone!!

- mobile: oh we need to fix that
- sorting: I think this is less of an issue, because I am keen to move the data to the API where it the order can be sorted there

eddiejaoude:
Now it is responsive 🎉 

Larger screens...

Screenshot 2021-09-21 at 09 33 06


Smaller screens ...

Screenshot 2021-09-21 at 09 33 14

src/Components/Milestones.js
```diff
@@ -0,0 +1,66 @@
...
+  )
+
+  const content = (milestone) => (
+    <Card title={milestone.title} subTitle={milestone.date} className="p-m-5">
+      {milestone.image && (
+        <img
+          src={milestone.image}
+          onError={(e) =>
+            (e.target.src =
+              'https://user-images.githubusercontent.com/624760/114314273-eaae0100-9af1-11eb-955a-4039657fe85a.png')
```
schmelto, line 30:
```suggestion
              'https://github.com/EddieHubCommunity.png')
```

schmelto (changes requested):
also the links are not working for me

eddiejaoude:
Going to merge if no more comments 🤓

PR #292, opened 2021-09-22
Update to .gitignore

### LINKED ISSUE #272: package-lock.json keeps being altered.
Every single pull request made by others contains a commit that shows the package-lock.json being altered. 
However, the user and creator of the pull-request hasn't done anything or made any changes. I think it automatically updates things such as the version every time someone forks and clones, and adds changes.

Thank you.

AbhijithGanesh:
Added package-lock.json to avoid sudden changes to dependencies.

Fixes #272 

## Proposed Changes

  - Added package-lock.json to avoid confusion(sudden dependency version change)
  -
  -

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,6 +2,8 @@
 
 # dependencies
 /node_modules
+.package-lock.json
+#Added this to avoid changes to Dependency files. 
 /.pnp
 .pnp.js
 


.gitignore
```diff
@@ -2,6 +2,8 @@
 
 # dependencies
 /node_modules
+.package-lock.json
+#Added this to avoid changes to Dependency files.
```
schmelto, line 6:
```suggestion
#Added this to avoid changes to Dependency files. 
.package-lock.json
```

for me the comments come before the line of code but this is only my personal preference

Panquesito7 (changes requested):
Please see the comment by @schmelto.

eddiejaoude:
The file is tracked by git, so the ignore will not do anything without the file being untracked by git - but the file is needed

Panquesito7:
> Thank you for the contribution.
> 
> The file is tracked by git, so the ignore will not do anything without the file being untracked by git - but the file is needed.
> 
> People should only commit files they have made changes too and not just blindly commit everything.

The file is needed, so if it's ignored by Git I don't think there will be any harm.

PR #418, opened 2021-10-02
Addition of newe Social Media platforms to the list

AbhijithGanesh:
Addition of the various social media to the list.


src/Components/Links.js
```diff
@@ -21,6 +21,10 @@ function Links({ links }) {
     twitter: '#00ACEE',
     vimeo: '#1AB7EA',
     youtube: '#FF0000',
+    hashnode: '#2962FF',
+    'dev.to': '#000000',
+    web: '#808080',
```
eddiejaoude, line 26: These are not currently support by the icon library, so we will have to park this PR for now

AbhijithGanesh: It's okay if this is parked for later merges but I added them to ensure we have more platforms

PR #465, opened 2021-10-06
Adding Skeleton to the User Profile when Loading

### LINKED ISSUE #443: [enhancement]Adding Skeleton while profile is loading
### Description

I think instead of ProgressBar there should be Skeleton when loading the profile
For reference : https://primefaces.org/primereact/showcase/#/skeleton

### Additional information

Changes to be made in 
1. https://github.com/EddieHubCommunity/LinkFree/blob/main/src/Components/Socials.js
2. https://github.com/EddieHubCommunity/LinkFree/blob/main/src/Components/Home.js

If the feature looks good i would like to make a pr

suhail34:
While profile of the user is loading it will show skeleton as it shows for chip

<a href="https://gitpod.io/#https://github.com/EddieHubCommunity/LinkFree/pull/465"></a>

closes #443

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,12 @@
+# [0.39.0](https://github.com/EddieHubCommunity/LinkFree/compare/v0.38.3...v0.39.0) (2021-10-24)
+
+
+### Features
+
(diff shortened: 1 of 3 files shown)


Panquesito7:
For reviewers: I'm just giving a link so you can see how this looks https://github.com/EddieHubCommunity/LinkFree/issues/443#issuecomment-936799035

PR #619, opened 2021-10-29
fix: more consistent loading skeleton

### LINKED ISSUE #601: [OTHER] Loading page and profiles page differences.

If you see the screenshots below, when the website is loading, it shows the first screenshot, website layout, but when it is loaded, all the profiles are aligned to the left, leaving a gap to the right. When it is loading, all the profiles are aligned to the center. 

The differences between the loading page and the profile page is the border of each profile, the alignment and how many of profile cards on each row.

## Here is when it is loading:
image

- More square borders.
- 6 profile cards on each row
- No gap to the right
- Aligned to center

## Here it is when it is loaded:

image

- Rounder borders.
- Aligned to left.
- Gap in the right.
- 4 profiles on each row.

I am not sure whether the profile page needs sorting or the loading page does. Thank you.

Vyvy-vi:
resolves #601 

## Description

This PR intends to reduce the differences between the loading skeleton and the Users page.
The `ProgressBar` on the main page has been removed since it overlapped with the skeleton for the `MenuBar`
(*though, I think maybe we can set the `ProgressBar` to appear before the Skeleton for the links gets rendered, in a later PR).

## Preview

https://user-images.githubusercontent.com/62864373/139380276-ac63d133-9a18-43a7-843c-45f36ab9bf03.mov

<a href="https://gitpod.io/#https://github.com/EddieHubCommunity/LinkFree/pull/619"></a>

### DIFF
--- a/src/Components/Home/Home.js
+++ b/src/Components/Home/Home.js
@@ -1,14 +1,12 @@
 import './Home.css'
 
 import React, { useState, useEffect, useRef } from 'react'
-import { ProgressBar } from 'primereact/progressbar'
 import { Toast } from 'primereact/toast'
 
 import Placeholders from './Placeholders'
 import Users from './Users'
 
 function Home() {
-  const [showProgress, setShowProgress] = useState(true)
   const [list, setList] = useState([])
   const [skeleton, setskeleton] = useState(true)
   const toast = useRef(null)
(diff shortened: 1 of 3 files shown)


Kunal0007 (approved):
LGTM 🎉

PR #642, opened 2021-11-05
fix: update Siddhi's bio

Panquesito7:
Things added/changed:

- Update Siddhi's bio from `Web development` to `Web developer`.

<a href="https://gitpod.io/#https://github.com/EddieHubCommunity/LinkFree/pull/642"></a>

### DIFF
--- a/public/data/SiddhiBhanushali.json
+++ b/public/data/SiddhiBhanushali.json
@@ -1,6 +1,6 @@
 {
   "name": "Siddhi Bhanushali",
-  "bio": "Web development & open source enthusiast",
+  "bio": "Web developer & open source enthusiast",
   "avatar": "https://github.com/siddhi-244.png",
   "links": [
     {


Kunal0007 (approved):

public/data/SiddhiBhanushali.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "Siddhi Bhanushali",
-  "bio": "Web development & open source enthusiast",
+  "bio": "Web developer & open source enthusiast",
```
eddiejaoude, line 3: I don't think we should update other people's profile - if that is what they put, then they should change it if they wish to do so

Panquesito7: > I don't think we should update other people's profile - if that is what they put, then they should change it if they wish to do so

Good idea 😅 closing this PR. Thank you, Eddie.

PR #660, opened 2021-11-12
Corrected element hierarchy

### LINKED ISSUE #650: heading heirarchy incorrect
### Description

On the individual profile page there is a h1 followed by an h4. A h1 should be followed by a h2 and size should be controlled by css to make the html semantically correct

### Screenshots

Screenshot 2021-11-09 143337

### Additional information

_No response_

SPARKY892:
resolve #650 

<a href="https://gitpod.io/#https://github.com/EddieHubCommunity/LinkFree/pull/660"></a>


ssensalo (approved):

mohan-murali:
I personally don't think name and username should be part of heading at all.  A screen reader is going to read it as heading and that is not really the heading of the page. It doesn't give correct meaning.

eddiejaoude:
@mohan-murali this is a page for the user (name) / username, I think it is the only relevant information we have for the page - we would want it specific rather than generic?

mohan-murali:
> @mohan-murali this is a page for the user (name) / username, I think it is the only relevant information we have for the page - we would want it specific rather than generic?

well, then should they both not be in the same heading hierarchy. Having username at a different heading hierarchy seems weird. I don't know, maybe its right. I am just thinking out loud here.

src/Components/Profile.js
```diff
@@ -26,7 +26,7 @@ function Profile({ profile, username }) {
           />
         </Avatar>
         <h1 className="p-m-2">{name}</h1>
```
krishguptadev, line 28:
```suggestion
        <h1 className="p-m-2">{name} <span class="p-m-2">({username})</span></h1>
```

src/Components/Profile.js
```diff
@@ -26,7 +26,7 @@ function Profile({ profile, username }) {
...
-        <h4 className="">({username})</h4>
+        <h2 className="p-avatar-username">({username})</h2>
```
krishguptadev, line 29:
```suggestion
```

krishguptadev (changes requested):

krishguptadev (approved):

Vyvy-vi (commented):
I think maybe the `name` and the `username` could be in different hierarchies?

krishguptadev:
> I think maybe the `name` and the `username` could be indifferent hierarchies?

I don't think that would be efficient for SEO and yes the issue this one is resolving.

src/Components/Profile.js
```diff
@@ -25,8 +25,7 @@ function Profile({ profile, username }) {
             className="p-mr-2 p-mt-2"
           />
         </Avatar>
-        <h1 className="p-m-2">{name}</h1>
-        <h4 className="">({username})</h4>
+        <h1 className="p-m-2">{name} <span class="p-m-2">({username})</span></h1>
```
krishguptadev, line 28:
```suggestion
        <div className="p-m-2 p-d-flex">
          <h1 className="p-m-0 p-r-3">{name}</h1>
          <h2 className="p-avatar-username">({username})</h2>
        </div>
```

krishguptadev (commented):
If we wanna implement different level headings for this as @Vyvy-vi suggested.

src/Components/Profile.js
krishguptadev: This also fixes the build, not sure about the tests though 😅

Vyvy-vi: moreover, this would also fix the failing test case(`react/no-unknown-property`)

SPARKY892: This unfortunately makes the username lower than the name
image

src/Components/Profile.js
```diff
@@ -25,8 +25,10 @@ function Profile({ profile, username }) {
             className="p-mr-2 p-mt-2"
           />
         </Avatar>
-        <h1 className="p-m-2">{name}</h1>
-        <h4 className="">({username})</h4>
+        <div className="p-m-2 p-d-flex">
+          <h1 className="p-m-0 p-r-3">{name}</h1>
```
krishguptadev, line 29: @SPARKY892 like this?

```suggestion
          <h1 className="p-m-2 p-r-3">{name}</h1>
```

src/Components/Profile.css
```diff
@@ -8,3 +8,7 @@
 .p-avatar-image > img {
   font-size: 0.5em;
 }
+
+.p-avatar-username {
```
emmalearnscode, line 12: Custom styles should not be needed.

src/Components/Profile.js
emmalearnscode, line 29: Using p-mr-2 would give the name a margin only to the right and stop them from being at different levels

src/Components/Profile.js
```diff
@@ -25,8 +25,10 @@ function Profile({ profile, username }) {
...
+          <h2 className="p-avatar-username">({username})</h2>
```
emmalearnscode, line 30: Please remove custom css

emmalearnscode (changes requested):
Please try to stick to prime react classes as much as possible. Just a couple of small changes and this will be ready to go :)

src/Components/Profile.js
```diff
@@ -25,8 +25,12 @@ function Profile({ profile, username }) {
             className="p-mr-2 p-mt-2"
           />
         </Avatar>
-        <h1 className="p-m-2">{name}</h1>
-        <h4 className="">({username})</h4>
+        <div className="p-m-2 p-d-flex">
+          <h1 className="p-m-0 p-r-3">{name}</h1>
+          <div className="p-m-2 p-d-flex">
+            <h1 className="p-m-0 p-r-3">{name}</h1>
```
emmalearnscode, line 31: There should only ever be one h1 per page due to accessibility reasons. So we either need an h1 and a h2 both name and username should be inside a h1 with an extra span to add different styles to the username.

emmalearnscode (changes requested):
F9037E4C-AD82-4416-84ED-51635EC1DC19

With the current styles these are being offset from each other and it doesn’t look great from a design perspective. Please try to style them so they sit on the same baseline.

src/Components/Profile.js
emmalearnscode: How's this coming along? Do you need any help @SPARKY892 ?

CodingSpecies (commented):
@SPARKY892 this pr has been open and not active for a while now. Do you need any help? Thank you.

emmalearnscode (approved):

krishguptadev (approved):

Vyvy-vi (approved):
This looks good to me :+1:

PR #661, opened 2021-11-13
fix: remove main tag in `User.js` causing duplicate

### LINKED ISSUE #651: Page can only have one main tag
### Description

On the profile page there is a main tag nested inside of another main tag when there should only be one.

### Screenshots

Screenshot 2021-11-09 143451

### Additional information

_No response_

Sfitzg:
Removed the main element in `User.js` as this was causing a duplicate main element

Closes #651 

<a href="https://gitpod.io/#https://github.com/EddieHubCommunity/LinkFree/pull/661"></a>

### DIFF
--- a/src/Components/UserProfile/ErrorPage.js
+++ b/src/Components/UserProfile/ErrorPage.js
@@ -4,10 +4,24 @@ function ErrorPage() {
   return (
     <div className="p-text-center">
       <div className="flex-column">
-        <img src='/eddiehub_community_logo.webp' alt="image" style={{ width: '150px' }}/>
+        <img
+          src="/eddiehub_community_logo.webp"
+          alt="image"
+          style={{ width: '150px' }}
+        />
         <h1>Profile not found.</h1>
         <h1>If you are a new user, please consider registering at LinkFree.</h1>
-        <h2>Read the documentation <a href="https://github.com/EddieHubCommunity/LinkFree#readme" target="_blank" rel="noreferrer">here</a>.</h2>
+        <h2>
+          Read the documentation{' '}
+          <a
+            href="https://github.com/EddieHubCommunity/LinkFree#readme"
+            target="_blank"
+            rel="noreferrer"
+          >
+            here
+          </a>
+          .
+        </h2>
       </div>
     </div>
   )
--- a/src/Components/UserProfile/User.js
+++ b/src/Components/UserProfile/User.js
@@ -26,7 +26,7 @@ function User() {
   }, [username])
 
   return (
-    <main>
+    <>
       <Navbar
         start={
           <Link to="/" aria-label="Go back to Home">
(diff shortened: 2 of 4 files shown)


mohan-murali (approved):

mohan-murali:
maybe you can try fix the failing test?

krishguptadev (changes requested):
Can you provide an ss of the inspector?

emmalearnscode (changes requested):
Looking at the structure of the code, I would suggest main be present in the home component and user component as these are the two main routes. Having main in too many child components complicates the structure and will lead to further problems later of main being duplicated again.

Sfitzg:
@emmalearnscode I agree with you but the structure was pretty confusing and I also feel that the main nav should be outside of the main area as it is consistent across pages?

krishguptadev (commented):
## Merge Conflict

Thank you for your contribution. Your pull request currently has conflicting changes with our `main` branch. In order to review this pull request, we will need those conflicts to be resolved.

We recommend doing this with the `git rebase` command. If you are unsure of how to fix this, let us know and we would be happy to assist you.

emmalearnscode:
> @emmalearnscode I agree with you but the structure was pretty confusing and I also feel that the site nav should be outside of the main area as it is consistent across pages?
> 
> I'm happy to amend but I also feel we should look at pulling the main navigation out of the main area? maybe in another issue ? Happy to hear your thoughts
> 
> thanks

Oh interesting, I hadn't noticed that the navbar was inside different components like that. Yes, ideally the navbar is outside of the routes so we should probably raise another issue.

src/Components/UserProfile/User.js
```diff
@@ -26,21 +26,23 @@ function User() {
...
         }
       />
-      {showProgress && <ProgressBar mode="indeterminate" />}
-      {skeleton && <Placeholder />}
-      {error && <ErrorPage />}
-      {!error && !skeleton && (
-        <ProfilePage profile={profile} username={username} />
-      )}
-    </main>
+      <main>
```
krishguptadev, line 37: Still a main tag here 😅

krishguptadev (changes requested):

src/Components/UserProfile/ErrorPage.js
```diff
@@ -4,10 +4,24 @@ function ErrorPage() {
...
+        <img
+          src="/eddiehub_community_logo.webp"
+          alt="image"
+          style={{ width: '150px' }}
+        />
         <h1>Profile not found.</h1>
         <h1>If you are a new user, please consider registering at LinkFree.</h1>
-        <h2>Read the documentation <a href="https://github.com/EddieHubCommunity/LinkFree#readme" target="_blank" rel="noreferrer">here</a>.</h2>
+        <h2>
+          Read the documendation{' '}
```
emmalearnscode, line 15: This reintroduces a spelling mistake in documentation.

src/Components/UserProfile/User.js
emmalearnscode: I understand why there is a main tag here, you are moving the navbar outside the main content which is correct. Can you take a screenshot of the outputted html code to make so we can see that it solves the issue and another main tag doesn’t sneak in from any other component.

Sfitzg:
Attached are the Screenshots 
Hope this is all in order now 😄 


## Home Page
Homepage

## Profile Page

UserProfile

## Error Page
Errorpage

krishguptadev (approved):

qwe123coder (approved):
LGTM! :rocket:

Vyvy-vi (approved):
Looks good to me!

Thank you for opening this PR, and contributing to LinkFree! :tada:

PR #662, opened 2021-11-13
ci: update workflow to work with protected branches

### LINKED ISSUE #659: [FEATURE] update releases to work with protected branches
### Description

Hello @eddiejaoude 

The current workflow for releases doesn't work with protected branches that require pull requests or branches that require status checks to pass before merging because the standard  `GITHUB_TOKEN` does not give admin access which is required to push to protected branches.

I have a workflow that will allow pushing to protected branches, it requires a Personal Access Token with admin rights to be set as a github secret to be able to achieve this. 
the workflow also checks if the `INCLUDE ADMINS` setting is set before doing the push and will temporarily disable `INCLUDE ADMINS` make the git push and then re enable `INCLUDE ADMINS`  if it was previously set.

I have setup the workflow in such a way that if the personal access token is not supplied in the github secrets it will revert to just using the previous workflow using the `GITHUB_TOKEN` therefore this workflow should work as before if the branch is not protected and admin token not supplied.

I would like to update the `releases.yml` to use this workflow!!

### Screenshots

_No response_

### Additional information

_No response_

Sfitzg:
Update the `releases.yml` workflow to work with protected branches by means of a environment variable being set with a personal access token containing admin rights.

closes #659 

<a href="https://gitpod.io/#https://github.com/EddieHubCommunity/LinkFree/pull/662"></a>

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -4,25 +4,78 @@ on:
     branches:
       - main
 
+env:
+  # For Use with protected branches must use a Personal Access Token with Admin rights
+  # For 'PERSONAL' repo must be repo 'OWNER' for 'ORGANISATION' can be any 'USER' with 'ADMIN' access
+  # change to either '' for standard GITHUB_TOKEN or ${{ secrets.YOUR_GITHUB_SECRET }} for personal access token
+  GH_TOKEN: ''
+
 jobs:
   changelog:
     runs-on: ubuntu-latest
 
     steps:
-      - uses: actions/checkout@v2
+      # If personal access token supplied then set it as GH_TOKEN
+      - name: 🔍 Using PERSONAL ACCESS TOKEN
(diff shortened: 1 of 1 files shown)


Panquesito7 (commented):
Is this a work in progress or done already? 🤔

Sfitzg:
@Panquesito7 it is complete but as I never received any confirmation on the issue  I just supplied a draft pull request so anyone could look at the changes.

Panquesito7 (commented):
How can we ensure this was tested properly? Was this ran in your fork?

krishguptadev (changes requested):
Check the latest comment on the issue.

Sfitzg:
I'm rather going to close this pull request as I don't think this is needed at this time.
The only real difference is that is disables the include admins checkbox and then re enables it after changelog is committed.

PR #669, opened 2021-11-15
Margin and Padding Decrease in Mobile View

### LINKED ISSUE #664: Less margin/padding on mobile screens to maximize screen area
### Description

There is a lot of padding round the links and cards on user profile page on mobile. This should be reduced to make better use of the space and means less text will be cutoff from the links

### Screenshots

Screenshot 2021-11-14 114724
Screenshot 2021-11-14 114741

### Additional information

_No response_

AbhinavMV:
Reduced Margin for max-screen size 768px
Closes #664 

After change view in mobile - 

image

<a href="https://gitpod.io/#https://github.com/EddieHubCommunity/LinkFree/pull/669"></a>

### DIFF
--- a/src/Components/Links.css
+++ b/src/Components/Links.css
@@ -30,3 +30,14 @@
   text-overflow: ellipsis;
   white-space: nowrap;
 }
+
+.w-70 {
+  width: 70%;
+  max-width: 45rem;
+}
+
+@media screen and (max-width: 768px) {
+  .w-70 {
+    width: 90%;
+  }
+}
--- a/src/Components/Links.js
+++ b/src/Components/Links.js
@@ -21,10 +21,7 @@ function Links({ links }) {
 
   return (
     <section className="p-d-flex p-jc-center p-mb-4">
-      <div
-        className="p-d-flex p-flex-column"
-        style={{ width: 70 + '%', maxWidth: 45 + 'rem' }}
-      >
+      <div className="p-d-flex p-flex-column w-70">
         {links
           .filter((link) => Object.keys(colors).includes(link.icon))
           .map((link, index) => (
@@ -38,7 +35,7 @@ function Links({ links }) {
                   MouseOver(e, colors[link.icon])
                 }}
                 onMouseOut={MouseOut}
-                className={`p-p-3 p-m-2 p-button-outlined ${link.icon}`}
+                className={`p-p-3 p-mt-2 p-mb-2 p-button-outlined ${link.icon}`}
                 style={{ color: colors[link.icon] }}
                 role="link"
               >
--- a/src/Components/Links.js
+++ b/src/Components/Links.js
@@ -35,7 +35,7 @@ function Links({ links }) {
                   MouseOver(e, colors[link.icon])
                 }}
                 onMouseOut={MouseOut}
-                className={`p-p-3 p-mt-2 p-mb-2 p-button-outlined ${link.icon}`}
+                className={`p-p-3 p-my-2 p-button-outlined ${link.icon}`}
                 style={{ color: colors[link.icon] }}
                 role="link"
               >
--- a/src/Components/Profile.css
+++ b/src/Components/Profile.css
@@ -13,7 +13,6 @@
   .w-50 {
     max-width: 90vw;
     margin: 2px auto;
-    text-align: center;
     line-height: 20px;
   }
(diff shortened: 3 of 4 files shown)


krishguptadev (approved):
LGTM!

Panquesito7 (approved):

src/Components/Links.js
```diff
@@ -38,7 +35,7 @@ function Links({ links }) {
                   MouseOver(e, colors[link.icon])
                 }}
                 onMouseOut={MouseOut}
-                className={`p-p-3 p-m-2 p-button-outlined ${link.icon}`}
+                className={`p-p-3 p-mt-2 p-mb-2 p-button-outlined ${link.icon}`}
```
mohan-murali, line 38: you can just add p-my-2 to achieve same

src/Components/Links.css
```diff
@@ -30,3 +30,14 @@
   text-overflow: ellipsis;
   white-space: nowrap;
 }
+
+.w-70 {
```
mohan-murali, line 34: Don't need this. Can be done with build in prime react classes
https://www.primefaces.org/primeflex/gridsystem
https://www.primefaces.org/primeflex/width

src/Components/Profile.css
```diff
@@ -8,3 +8,12 @@
 .p-avatar-image > img {
   font-size: 0.5em;
 }
+
+@media screen and (max-width: 768px) {
```
mohan-murali, line 12: This can be done with built in prime react classes
https://www.primefaces.org/primeflex/width
https://www.primefaces.org/primeflex/lineheight
https://www.primefaces.org/primeflex/textalign

mohan-murali: I just noted that we can't change all properties within breakpoints with prime flex version 2.

src/Components/Profile.css
```diff
@@ -8,3 +8,12 @@
...
+  .w-50 {
+    max-width: 90vw;
+    margin: 2px auto;
+    text-align: center;
```
mohan-murali, line 16: This is same even outside media query

src/Components/Links.js
AbhinavMV: Sure will do this change 👍

src/Components/Profile.css
AbhinavMV: so will keep this media query then?

src/Components/Profile.css
AbhinavMV: just noticed it. will remove the text-align property.

src/Components/Links.css
AbhinavMV: just tried using p-grid and p-col 
width is not getting adjusted properly based on screen size extra class is needed for this.
After changing to grid style -> (need to limit the max size based on responsiveness)
image

Also could you tell me why prime react class would be better choice instead of normal css classes?

src/Components/Profile.css
AbhinavMV: Done 👍

src/Components/Links.js
AbhinavMV: Done 👍

Pradumnasaraf (approved):
LGTM 🦅

src/Components/Profile.css
mohan-murali: Yea keep it. Its fine.

src/Components/Links.css
mohan-murali: I guess its fine. Prime react doesn't seem to have % based widths. And its always better to use a css library classes that we are already using in our project instead of adding css on our own. CSS is hard to maintain and having duplicate CSS only increases the problem. Also we will be shipping extra css into the client that uses this app, which is never good.

mohan-murali (approved):

krishguptadev (approved):

shreyaansjain06 (approved):
LGTM

src/Components/Links.css
AbhinavMV: oohh ok. 
Thanks for letting me know. Will keep this in mind. 😄

qwe123coder (approved):
LGTM :rocket:

Vyvy-vi (approved):
Thanks for contributing to LinkFree!

PR #672, opened 2021-11-16
chore: update .gitpod.yml removing the link to the dockerfile

### LINKED ISSUE #671: [OTHER] Gitpod does not need its own Dockerfile

I noticed when quickly checking out the project with Gitpod that it was taking quite some time to load and this is due to the necessity of Gitpod needing to do additional tasks (like install npm) but also not be able to use caching for the workspace.

### Here's the low-down:
1. the `gitpod/workspace-full` already has npm installed
2. you can use `.gitpod.yml` tasks to do things inside the box if necessary

### Proposed fix
1. remove Dockerfile
2. find out what that breaks
3. add appropriate run tasks to `.gitpod.yml`

stemount:
Fixes #671

### Steps to test
1. Gitpod cold start (e.g. new workspace) should be super speedy
2. Ensure everything still plays nicely.

<a href="https://gitpod.io/#https://github.com/EddieHubCommunity/LinkFree/pull/672"></a>

### DIFF
--- a/.gitpod.dockerfile
+++ b/.gitpod.dockerfile
@@ -1,3 +1,0 @@
-FROM gitpod/workspace-full
-
-RUN npm i -g npm
--- a/.gitpod.yml
+++ b/.gitpod.yml
@@ -18,6 +18,3 @@ github:
     addCheck: true
     addComment: false
     addBadge: true
-
-image:
-  file: .gitpod.dockerfile


Panquesito7 (approved):
LGTM! 🚀

stemount:
wait turns out @Cahllagerfeld may have a reason for this as I noticed https://github.com/EddieHubCommunity/api/blob/main/.gitpod.dockerfile allegedly this speeds up things?

Out of curiosity, is this a trick that I have missed?

My assumptions are:
1. `npm` inside the Gitpod workspace-full should be reasonably recent so it is not this
2. It could _possibly_ be faster if a project needs a later version of NPM and significant speed improvements happen here?
3. It is faster to boot for some reason *only* if you have Gitpod pre-builds?

Share the method to the madness as I am guessing blindly 🙈

krishguptadev (approved):
I think these changes are good.

Cahllagerfeld:
- Prebuilds boost the spin-up significantly
- this dockerfile comes from a time where npm switch between lockfile-version 1 and 2. The Gitpod-base-image had the old npm version back then which caused, that the generated lock-file when installing it always was the old format although the new was already out

stemount:
@Cahllagerfeld then what do you think about this PR?

Usually, the gitpod workspace should be quicker as prebuilds aren't always going to be available to all users - unless I've misunderstood it, it is not actually used unless you add something as a "project" right?

I think you may know more than me here RE: this improvement of upgrading npm in the base image (?). I found it to be more performant without it but that was with cold starts.

##### the npm v6 vs v7 problem
yeah I saw this becoming a problem specifically in LinkFree especially with folks recommitting in package-lock.json as v2, although I would say most are actually getting the V2 lockfile (it's supported in V6) and NPM v7 has caused many headaches for users of frontend frameworks for reasons I have forgot.

Whatcha think?

shreyaansjain06 (approved):
LGTM

Cahllagerfeld:
Prebuilds are configured for every user, but per project. in the `gipod.yml` you can define in which scenarios prebuilds are made.

I think we can remove the docker-image though, because the npm-version-bump is done afaik in the base image now

Cahllagerfeld (approved):

Panquesito7 (approved):

krishguptadev (approved):

Panquesito7:
I'll merge this as it already has approval from Cahl. 😄

PR #682, opened 2021-11-19
Name Initials as avatar for cases where image does not exist (fixes #667)

### LINKED ISSUE #667: Add Default Profile Image
### Description

If an image is not visible we can show a default image with Initial character from name (eg: Like it is shown on gmail)

### Screenshots

image
image

### Additional information

_No response_

harshrai654:
Used [dicebear](https://www.npmjs.com/package/@dicebear/avatars) To produce SVG of name initials to be replaced on Event `onImageError` in ```Chip``` and ```Avatar``` Component.

Screenshot from 2021-11-20 00-04-14

Screenshot from 2021-11-20 00-04-32

<a href="https://gitpod.io/#https://github.com/EddieHubCommunity/LinkFree/pull/682"></a>

resolve #667


harshrai654:
fixes #667

public/data/harshrai654.json
```diff
@@ -0,0 +1,12 @@
+{
```
krishguptadev, line 1: Can you move this to a different PR 😅

krishguptadev (changes requested):

public/data/harshrai654.json
harshrai654: Ok, so should i remove my profile.json file from this PR?

krishguptadev: Yep

src/Components/Home/Users.js
```diff
@@ -36,7 +38,20 @@ function Users({ list }) {
...
+              <Chip
+                image={user.avatar}
+                className="p-m-2"
+                label={user.name}
+                onImageError={(error) => {
+                  const defaultSVG = createAvatar(style, {
+                    seed: user.name,
+                    dataUri: true,
+                  })
+
```
mohan-murali, line 50:
```suggestion

```

src/Components/Profile.js
```diff
@@ -18,6 +20,15 @@ function Profile({ profile, username }) {
           size="xlarge"
           shape="circle"
           className="p-overlay-badge"
+          onImageError={(error) => {
+            const defaultSVG = createAvatar(style, {
+              seed: name,
+              dataUri: true,
+            })
+
```
mohan-murali, line 28:
```suggestion

```

src/Components/Home/Users.js
krishguptadev: I do not think this line needs to be removed (it just makes easier to read)

src/Components/Profile.js
krishguptadev: same here.

krishguptadev (changes requested):

src/Components/Home/Users.js
```diff
@@ -36,7 +38,20 @@ function Users({ list }) {
           filteredList.length > 0 &&
           filteredList.map((user, key) => (
             <Link to={user.username} key={`avatar-${key}`}>
-              <Chip image={user.avatar} className="p-m-2" label={user.name} />
+              <Chip
+                image={user.avatar}
+                className="p-m-2"
+                label={user.name}
+                onImageError={(error) => {
```
krishguptadev, line 45: Since we use this function twice, how about creating a helper function for this.

harshrai654: yeah, that would be good, any common path where such function can be moved?

harshrai654: I am not able to find any common file where all such utility functions can be placed, how about creating a `src/utils.js` file? From there we can import all such functions.

krishguptadev: Yep! Feel free to create src/utils.js

harshrai654: I have created a helper function for default image behavior.

krishguptadev (approved):

harshrai654:
Is this PR ready for merge?

krishguptadev:
> Is this PR ready for merge?

We need to wait for a maintainer to review 🙂

Vyvy-vi (changes requested):
It looks like the `package-lock.json` was deleted from the fork.
It would be great if you could `reset` that file

krishguptadev (commented):
@Vyvy-vi doesn't seem to me 😅 https://github.com/harshrai654/LinkFree/blob/issue-667/package-lock.json

Vyvy-vi:
Oh, right 😅 
It isn't removed, it has a few deletions-
`184 additions, 26,722 deletions not shown because the diff is too large. Please use a local Git client to view these changes.`

I'm wondering, what might have caused these deletions?

krishguptadev:
> Oh, right 😅 It isn't removed, it has a few deletions- `184 additions, 26,722 deletions not shown because the diff is too large. Please use a local Git client to view these changes.`
> 
> I'm wondering, what might have caused these deletions?

maybe they converted package-lock file version 🤔

krishguptadev:
Can be fixed by running `npm install`

harshrai654:
> Can be fixed by running `npm install`

yes, this should fix it.

krishguptadev:
Made PR to fix it, don't know why but I had to make the branch up to date before making this PR: https://github.com/harshrai654/LinkFree/pull/1

schmelto:
Can we also achieve this by using prime react components? I saw that the avatar in primereact can use labels to achieve the initials. Not sure if this so already implemented in the chips using here.

harshrai654:
> Can we also achieve this by using prime react components? I saw that the avatar in primereact can use labels to achieve the initials. Not sure if this so already implemented in the chips using here.

Label prop in Avatar will give something like this:
Screenshot from 2021-11-22 15-06-01

and I am not sure if it is possible to set a label prop in `onImageError` event callback since we want initials only when there is an error in fetching the image.

AbhinavMV:
> > Can we also achieve this by using prime react components? I saw that the avatar in primereact can use labels to achieve the initials. Not sure if this so already implemented in the chips using here.
> 
> Label prop in Avatar will give something like this: Screenshot from 2021-11-22 15-06-01
> 
> and I am not sure if it is possible to set a label prop in `onImageError` event callback since we want initials only when there is an error in fetching the image.

I think you can do this ->
create a new state variable
onImageError  set this state to true
then we can load image based on this state
eg -> newImageErrorState? <Avatar label/> : 

Avatar is present in primereact. This way we won'nt need a new library. 
I am not sure if this is a better method maybe someone can confirm this. 🙂

harshrai654:
> > > Can we also achieve this by using prime react components? I saw that the avatar in primereact can use labels to achieve the initials. Not sure if this so already implemented in the chips using here.
> > 
> > Label prop in Avatar will give something like this: Screenshot from 2021-11-22 15-06-01
> > 
> > and I am not sure if it is possible to set a label prop in `onImageError` event callback since we want initials only when there is an error in fetching the image.
> 
> I think you can do this ->
> create a new state variable
> onImageError  set this state to true
> then we can load image based on this state
> eg -> newImageErrorState? < Avatar label /> : < img />
> 
> Avatar is present in primereact. This way we won'nt need a new library. 
> I am not sure if this is a better method maybe someone can confirm this. 🙂
> 
Conditional rendering can be used, but as i have shown in above image, label does not provide the styling needed, dicebear on the other hand provides stylings for all use cases which can also be extended for future features.

AbhinavMV:
> > > > Can we also achieve this by using prime react components? I saw that the avatar in primereact can use labels to achieve the initials. Not sure if this so already implemented in the chips using here.
> > > 
> > > 
> > > Label prop in Avatar will give something like this: Screenshot from 2021-11-22 15-06-01
> > > and I am not sure if it is possible to set a label prop in `onImageError` event callback since we want initials only when there is an error in fetching the image.
> > 
> > 
> > I think you can do this ->
> > create a new state variable
> > onImageError  set this state to true
> > then we can load image based on this state
> > eg -> newImageErrorState? < Avatar label /> : < img />
> > Avatar is present in primereact. This way we won'nt need a new library.
> > I am not sure if this is a better method maybe someone can confirm this. 🙂
> 
> Conditional rendering can be used, but as i have shown in above image, label does not provide the styling needed, dicebear on the other hand provides stylings for all use cases which can also be extended for future features.

To achieve that styling you can use other attributes like shape badge.. https://www.primefaces.org/primereact/showcase/#/avatar
A demo on the website shows it can be achieved. 😄

harshrai654:
I have tried to use Avatar Component's `label` prop:
Screenshot from 2021-11-23 12-35-53

Chip component's label prop is for text usage, so its behaviour is different, the only way I see is to replace the image with SVG image of name initial if anyone knows a way to give text-based name initial in Chip Componet it will be helpful as it will remove `dicebear` dependency.

emmalearnscode:
This seems like an awful lot of code and extra dependencies to achieve this and I wonder if maybe a default profile icon would be better to use as the name and username are printed beside the avatar anyway so the initials are not really necessary.

emmalearnscode:
@harshrai654 I see also that someone else already requested to work on this issue. Now if you are both working on it someone’s time and effort will be wasted and people will get upset. In the spirit of fairness please discuss in the issue with the original person that asked before submitting a pr. if they are still working on it then they should be left to it, if they no longer want the issue then you can ask to be assigned instead.

FlorianKoerner:
As an alternative, you could just use the API.

Example:
https://avatars.dicebear.com/api/initials/Florian%20Koerner.svg

Documentation:
https://avatars.dicebear.com/docs/http-api

harshrai654:
> @harshrai654 I see also that someone else already requested to work on this issue. Now if you are both working on it someone’s time and effort will be wasted and people will get upset. In the spirit of fairness please discuss in the issue with the original person that asked before submitting a pr. if they are still working on it then they should be left to it, if they no longer want the issue then you can ask to be assigned instead. 

Okay, sure let me calrify this issue status with original person.

harshrai654:
> As an alternative, you could just use the API.
> 
> Example:
> https://avatars.dicebear.com/api/initials/Florian%20Koerner.svg
> 
> Documentation:
> https://avatars.dicebear.com/docs/http-api

Yes, that is an option, but considering a scenario where error occurs in most of the images in home page, this will cause many API request which may make the page load slow.

FlorianKoerner:
> Yes, that is an option, but considering a scenario where error occurs in most of the images in home page, this will cause many API request which may make the page load slow.

I think this is a very theoretical scenario. This should not really happen in this form.

And even if it does, the API supports HTTP2 and is provided via a CDN. That should be enough for a fallback.

And perhaps more importantly, with the API you don't load additional dependencies when you don't have to in the case of success.

harshrai654:
> > Yes, that is an option, but considering a scenario where error occurs in most of the images in home page, this will cause many API request which may make the page load slow.
> 
> I think this is a very theoretical scenario. This should not really happen in this form.
> 
> And even if it does, the API supports HTTP2 and is provided via a CDN. That should be enough for a fallback.
> 
> And perhaps more importantly, with the API you don't load additional dependencies when you don't have to in the case of success.

Yes dependency overhead is bad, API request will be better, but now a decision is to be taken regrading we need initials or a single default profile image.

Vyvy-vi:
I think it can be neat to have initials.
Many existing applications do the same, or they generate pseudo-random avatars for people...

harshrai654:
> I think it can be neat to have initials.
> Many existing applications do the same, or they generate pseudo-random avatars for people...

Yeah, its look simple and good.

harshrai654:
Removed Dicebear as a dependency, using Dicebear's API as fallback default profile image as username's initials.

harshrai654:
> Removed Dicebear as a dependency, using Dicebear's API as fallback default profile image as username's initials.

@emmalearnscode can you please review the changes?

src/utils.js
```diff
@@ -0,0 +1,7 @@
+export default {
+  setDefaultSVG: async function(name, error) {
+    const url = `https://avatars.dicebear.com/api/initials/${name}.svg`
```
emmalearnscode, line 3: Does this line need an await?

emmalearnscode (changes requested):
If nothing now changes in package.json and package-lock then they should not be included in the pull request

src/utils.js
harshrai654: Nope, its just a simple url for the image `src` attribute, it is not doing any fetch.

harshrai654:
> If nothing now changes in package.json and package-lock then they should not be included in the pull request 

`dicebear/avatar` package is no longer needed so it is removed from package.json.

src/utils.js
harshrai654: `async` is not needed here so i have removed it.

Vyvy-vi (changes requested):
Perhaps, you can `reset` the `package-lock.json` file, so that it's back to the earlier state?

(you can do so by using the command - `git reset package-lock.json origin/main`)

Vyvy-vi (changes requested):
This looks good to me!

However, there is an issue here.
The initials act as a fallback when the wrong link is provided, however if the `avatar` key is not present in the `<username>.json` file, the default icon made from initials doesn't show up.

(but that can always be something that's tackled in a later PR)

Vyvy-vi (approved):
LGTM!
We can ship this for now :)

Thanks for contributing to LinkFree!

krishguptadev (approved):

PR #703, opened 2021-11-25
feat(add-profile): EddieHubCommunity

krishguptadev:
RR: @eddiejaoude 

<a href="https://gitpod.io/#https://github.com/EddieHubCommunity/LinkFree/pull/703"></a>

Issue #696, opened 2021-11-23
Add GitHub Organisation profiles suggestion.


CodingSpecies:
### Description

I just thought as an idea if it was possible for us to submit profiles of our organisations in LinkFree.
- Maybe, in the bio / description we can explain how to join the org/ what the org is about.
- However, this can not be any organisation, but only the ones you actual are the founder of.
- There could be a seperate page to view different organisations, or in the same page, but highlighted differently so it differs from user profiles.
Thank you.

I am not sure if this can be implemented, but I just thought as an idea.

krishguptadev:
I like the idea, but don't support them to be displayed in a highlighted way.

CodingSpecies:
I was thinking as the user profiles are rounded, maybe the organisation profiles should have some different styling to perhaps look different to the user profiles. This is so you can see which ones are clearly organisations and what are not. @krishguptadev

krishguptadev:
> I was thinking as the user profiles are styled maybe the organisation profiles should have some different styling to perhaps look different to the user profiles. This is so you can see which ones are clearly organisations and what are not. @krishguptadev

But I don't think we need this. Organisation Profiles should be like normal profiles.

CodingSpecies:
Yes @krishguptadev i don’t think it is necessary now. Thank you

krishguptadev:
Can the issue be closed for the same?

CodingSpecies:
I don’t know because I don’t have the answer if we  are able to add the organisations. Once I have the answer, then this can be closed @krishguptadev

krishguptadev:
> I don’t know because I don’t have the answer if we are able to add the organisations. Once I have the answer, then this can be closed @krishguptadev

I just added a org, to clear the doubt 😄 #703

CodingSpecies:
So now at LinkFree we can add our own organisation profiles, shouldn't this be made clear on the CONTRIBUTING.md or README.md?

CodingSpecies:
As I said in #704 
To make only organisations come up in the search, maybe at the end of each profile name, the emoji 📛, which is the badge could be put. `EG: Eddie Hub 📛`.
When you put 📛 in the search all the organisations will come up, as they all have the same badge emoji in their name.

krishguptadev:
[...]

PR #704, opened 2021-11-25
feat(add-profile): web3community

krishguptadev:
Cc: @vinzvinci 

<a href="https://gitpod.io/#https://github.com/EddieHubCommunity/LinkFree/pull/704"></a>

### DIFF
--- a/public/data/web3community.json
+++ b/public/data/web3community.json
@@ -14,7 +14,7 @@
       "icon": "book"
     },
     {
-      "name": "Mail us",
+      "name": "Contact us via mail",
       "url": "mailto:web3community@gmail.com",
       "icon": "envelope"
     },
--- a/public/data/web3community.json
+++ b/public/data/web3community.json
@@ -1,0 +1,48 @@
+{
+  "name": "Web3 Community",
+  "bio": "An Open Source Community that focuses on decentralized applications, web 3, and blockchain technologies 🚀",
+  "avatar": "https://github.com/web3community.png",
+  "links": [
+    {
+      "name": "Visit Website",
+      "url": "https://web3community.github.io",
+      "icon": "globe"
+    },
+    {
+      "name": "Read our Blog",
+      "url": "https://dev.to/web3community",
+      "icon": "book"
+    },
+    {
+      "name": "Mail us",
+      "url": "mailto:web3community@gmail.com",
+      "icon": "envelope"
+    },
+    {
+      "name": "Visit GitHub Organisation",
+      "url": "https://github.com/web3community",
+      "icon": "github"
+    },
+    {
+      "name": "Join our Discord Community",
+      "url": "https://discord.gg/TSRwqx4K2v",
+      "icon": "discord"
+    },
+    {
+      "name": "Follow us on Twitter",
+      "url": "https://twitter.com/web3community",
+      "icon": "twitter"
+    },
+    {
+      "name": "LinkedIn Organisation",
+      "url": "https://www.linkedin.com/company/web3community",
+      "icon": "linkedin"
+    },
+    {
+      "name": "Check out our Instagram",
+      "url": "https://www.instagram.com/web3community",
+      "icon": "instagram"
+    }
+  ],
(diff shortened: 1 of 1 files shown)


vinzvinci (commented):
LGTM! Is org. allow to add?

krishguptadev:
> LGTM! Is org. allow to add?

Yep! I guess!

Panquesito7 (commented):
I'm not sure if organizations are allowed to add 😅 @EddieHubCommunity/linkfree, thoughts?

Panquesito7:
Actually, I think yes, it's allowed (see #703).

public/data/web3community.json
```diff
@@ -0,0 +1,48 @@
+{
+  "name": "Web3 Community",
+  "bio": "An Open Source Community that focuses on decentralized applications, web 3, and blockchain technologies 🚀",
```
Panquesito7, line 3:
```suggestion
  "bio": "An Open Source Community that focuses on decentralized applications, Web 3.0, and Blockchain technologies 🚀",
```

public/data/web3community.json
```diff
@@ -0,0 +1,48 @@
...
+      "url": "https://web3community.github.io",
+      "icon": "globe"
+    },
+    {
+      "name": "Read our Blog",
+      "url": "https://dev.to/web3community",
+      "icon": "book"
+    },
+    {
+      "name": "Mail us",
```
Panquesito7, line 17: Optional, just an idea. 😃
```suggestion
      "name": "Contact us via mail",
```

public/data/web3community.json
krishguptadev: It should lowercase as per grammarly.

Panquesito7: > It should lowercase as per grammarly.

Not for my Grammarly settings, though 👀 let's keep it as-is anyways.
But as you can see on the website we're using `Web 3.0`.

image

krishguptadev: As you can see in the github org bio, we are using web3 🙂

starlightknown (changes requested):
Maybe make it follow us on LinkedIn than just LinkedIn organization? Otherwise it looks good to me

krishguptadev:
> Maybe make it follow us on LinkedIn than just LinkedIn organization? Otherwise it looks good to me

Can we follow linkedin orgs? Sorry, I am not on LinkedIn, thus don't know.

Vyvy-vi:
Maybe we should add a tag in the profile schema, that declares an org as an 'org'...

krishguptadev:
> Maybe we should add a tag in the profile schema, that declares an org as an 'org'...

#696 🤔

eddiejaoude:
I think they are a community and open source, then it could give the project some exposure too - this might change in the future, it is just my initial thinking

emmalearnscode:
I think having some way to differentiate organisations from individuals would be a good idea. I might want to search just for orgs for example.

Cahllagerfeld:
> I think they are a community and open source, then it could give the project some exposure too - this might change in the future, it is just my initial thinking

I think especially when the new partners program is ready. The partners should have benefits in comparision to non-partners I think. One might be to be listed in this project or something

krishguptadev:
Maybe we shall refer #696 for this discussion 😅

Panquesito7 (approved):

CodingSpecies:
> I think having some way to differentiate organisations from individuals would be a good idea. I might want to search just for orgs for example.

Yes, I agree with that @emmalearnscode .
I was thinking that there should be a clear difference between orgs and user profiles. Maybe at the end of each profile name, the emoji 📛, which is the badge could be put. `EG: Eddie Hub 📛`.
When you put 📛 in the search all the organisations will come up as they all have the same badge emoji in their name.

eddiejaoude:
Great idea to differentiate 👍 , what about a field to do this - then we could have drop down or radio buttons to filter

avneesh0612:
A dropdown would be cool 🤔👀

PR #366, opened 2021-10-01
Social icons

mohammad4kh:
Adding different social media icons to the site

For example

 - Facebook
 - Instagram
 - Linkedin
 - Vimeo
 - Discord

and others 

Screenshot from 2021-10-01 02-33-57

<a href="https://gitpod.io/#https://github.com/EddieHubCommunity/LinkFree/pull/366"></a>

Resolve #554

Issue #106, opened 2021-09-13
Add more links like linkedIn, Instagram in the LinkFree


Pulkit007:
We should add support for additional social platforms such as:

- LinkedIn
- Instagram
- Discord

Each social should have support for the platform's icon.

StarTrooper08:
Dev.to forem link can be added can I add it??

Soham041201:
Anybody working on this?

StarTrooper08:
yep I m working to add devto forem

Omkar845:
i would like to contribute

Sejal-16:
can i work on this ?

StarTrooper08:
Discord link PR is ready and also ready to merge I guess.
And Instagram link is added .
Any other social media link in your mind and if you are Confused how to make first open source contribution I m there to help.

Tag me ( @StarTrooper08 )

StarTrooper08:
> Can i add discord link

LinkedIn link is remaining I guess check for the closed PR.
If no one added linkedin you can add it😃

Zishan-7:
Hey @Vyvy-vi , I see that instagram and Linkedin links haven't been added yet. Can you assign me this issue?

krishguptadev:
> Hey @Vyvy-vi , I see that instagram and Linkedin links haven't been added yet. Can you assign me this issue?

#675 doesn't this add the LinkedIn icon

Zishan-7:
> > Hey @Vyvy-vi , I see that instagram and Linkedin links haven't been added yet. Can you assign me this issue?
> 
> #675 doesn't this add the LinkedIn icon

No one has added Linkedin icon yet.

krishguptadev:
> > > Hey @Vyvy-vi , I see that instagram and Linkedin links haven't been added yet. Can you assign me this issue?
> > 
> > 
> > #675 doesn't this add the LinkedIn icon
> 
> No one has added Linkedin icon yet.

But there is a open PR for this 😅 Let's wait for the maintainers

Zishan-7:
> > > > Hey @Vyvy-vi , I see that instagram and Linkedin links haven't been added yet. Can you assign me this issue?
> > > 
> > > 
> > > #675 doesn't this add the LinkedIn icon
> > 
> > 
> > No one has added Linkedin icon yet.
> 
> But there is a open PR for this 😅 Let's wait for the maintainers

yea, sure, 😄

vr513:
[...]

PR #728, opened 2021-11-27
fix: LinkedIn icon

qwe123coder:


krishguptadev (changes requested):
I think this has to be linkedin, because we will be adding the icon

qwe123coder:
> I think this has to be linkedin, because we will be adding the icon

it would take a while to make a pr and that getting merged so i was thinking as of now we can keep it as global
 after you folks add linkedin logo i can make a pr to change it back to the same

qwe123coder:
ooo found this pr #366 if we merg that then i think the problem will be solved

Pradumnasaraf (commented):
> Yes, it will take a little while to add the LinkedIn logo.

**I don't think we need to change it to `globe`. Because after addition (LinkedIn Logo), we have to again change it back to `LinkedIn`. So, doesn't sound lucrative to me.**

Love to hear other reviewers thoughts on this :)

emmalearnscode:
I agree with @Pradumnasaraf , the icon pack is currently being discussed and this change would just cause problems later when the LinkedIn icon becomes available.

krishguptadev:
btw github provides a single button to revert a whole PR: 
image

Panquesito7 (commented):
As discussed above, I think it's best to wait until a change is made. 🙂

krishguptadev (commented):
+1 to @Panquesito7's comment

qwe123coder:
> +1 to @Panquesito7's comment

cool

Vyvy-vi:
> btw github provides a single button to revert a whole PR: image

Even though the button exists, our project has been moving quite fast and leaving PRs around for a while makes it tough for the original contributor as well as the maintainers, since the project changes too many aspects, and thus the PRs original purpose gets lost...

Vyvy-vi (commented):
Thanks for contributing to Linkfree!
While this addition _can_ add some value, I don't think this is needed especially considering that there's a very extended discussion on icons going on, and this is going to be one of the changes that are going to stick around if merged in, and would probably be a lot of manual work to edit...

This does remind us that we should probably resolve the icons discussion soon...

Vyvy-vi:
While this PR is being closed, we really appreciate that you took the time to go through all the files and edit them(I certainly hope that it was a script, rather than something done manually 😬 ). Though, in these cases, I think it could have benefited everyone if this was discussed on an issue like #106 or a new one.

Feel free to check out other issues on LinkFree, if they interest you!

PR #820, opened 2021-12-16
feat: add customizable profile background theme

Tejas-Teju:
## Changes proposed
Now, users can view the background theme of user's choice when they visit profile page however, the background theme must be present in <profile_name>.json file else default background is set i.e. #fff (White). 

## Screenshots
image
image
image

<a href="https://gitpod.io/#https://github.com/EddieHubCommunity/LinkFree/pull/820"></a>

### DIFF
--- a/public/data/Tejas-Teju.json
+++ b/public/data/Tejas-Teju.json
@@ -1,6 +1,7 @@
 {
     "name": "Tejas S",
     "type": "personal",
+    "theme": "#ffe680",
     "bio": "A Coding Enthusiast ",
     "avatar": "https://github.com/Tejas-Teju.png",
     "links": [
--- a/src/Components/Home/Home.js
+++ b/src/Components/Home/Home.js
@@ -36,6 +36,7 @@ function Home() {
       })
   }, [])
 
+  document.body.style.background = '#fff'
   return (
     <main>
       <Toast ref={toast} />
--- a/src/Components/UserProfile/ProfilePage.js
+++ b/src/Components/UserProfile/ProfilePage.js
@@ -6,6 +6,7 @@ import Links from '../Links'
 import Milestones from '../Milestones'
 
 function ProfilePage({ profile, username }) {
+  document.body.style.background = profile.theme
   return (
     <>
       {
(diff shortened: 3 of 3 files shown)


emmalearnscode:
This is going to create massive accessibility issues as the contrasts won’t always be up to wcag standards. I wonder if we can have a white panel in the middle and customisable bars down the sides (if we really must have customisable colours at all, there’s already a lot of colours with different icons and buttons, this feels like overkill but that is just my opinion)

Vyvy-vi:
I think eventually, we would want to move away from custom colors based on platforms (for profiles using themes besides the default Linkfree theme), and make the themes a collection of button colours, text colours and background colours?

Tejas-Teju:
> This is going to create massive accessibility issues as the contrasts won’t always be up to wcag standards. I wonder if we can have a white panel in the middle and customisable bars down the sides (if we really must have customisable colours at all, there’s already a lot of colours with different icons and buttons, this feels like overkill but that is just my opinion)

Yes @emmalearnscode you are absolutely right but this is just a MVP version of it and as said by @Vyvy-vi we'll enhance the feature going forward.

emmalearnscode:
> > This is going to create massive accessibility issues as the contrasts won’t always be up to wcag standards. I wonder if we can have a white panel in the middle and customisable bars down the sides (if we really must have customisable colours at all, there’s already a lot of colours with different icons and buttons, this feels like overkill but that is just my opinion)
> 
> Yes @emmalearnscode you are absolutely right but this is just a MVP version of it and as said by @Vyvy-vi we'll enhance the feature going forward.

I'm not sure this is an MVP though, it's going to create problems with accessibility and therefore not viable. I think that really, instead of allowing random background colors, we need to make a plan for the things that @Vyvy-vi mentions above about themes, then work out how that will be implemented and create an MVP from that.

Vyvy-vi:
Perhaps, we can incolde the themeing code, but ignore and not document the theme key in the data files?

emmalearnscode:
> Perhaps, we can include the themeing code, but ignore and not document the theme key in the data files?

I'm not sure that's a great idea. It could still be used by some people and create accessibility issues. And if we decide to take a different approach to themes when it's been properly discussed then we will need to go through and remove it from the profiles that have included it.

Cahllagerfeld:
I'm the same opinion as @emmalearnscode here. This might cause major problems with a11y. We should make a concept (if we want custom theming and if yes how we want to realize that to keep a11y high) first 🤔

Tejas-Teju:
Same opinion as @emmalearnscode, for now we can drop this PR and discuss more about the customization If everyone agrees.

emmalearnscode:
Yes, let’s move the discussion back to issues for now and maybe do some mock ups before proceeding with design changes

PR #850, opened 2021-12-23
fix: avatar aspect ratio distortion

### LINKED ISSUE #847: Names that are too long , Not getting wrapped properly
### Description

As you can see on the below screen shot.

The names that are too long , causing a problem of not getting wrapped properly which distorting the avatar image.

To solve this :
1. The layout Should be rows instead of columns in small devices
2. We should add character limit for the names.

The 2nd point is not necessary for now ,
It will be implemented when data will be taken as a input via a form UI  with authentication service.

### Screenshots

Screenshot_2021-12-23-11-19-58-096_com android chrome

### Additional information

_No response_

nathgoutam93:
## Fixes Issue

Closes #847 

## Changes proposed

- [X] without a fix width and height, Avatar's image height and width properties were getting overwritten
- [X] fixed it via giving it a fix width and height value which is equal to the avatar size = "xlarge"

## Screenshots
before fix:
Screenshot 2021-12-23 at 18-04-21 LinkFree
after fix:
Screenshot 2021-12-23 at 18-04-02 LinkFree

<a href="https://gitpod.io/#https://github.com/EddieHubCommunity/LinkFree/pull/850"></a>


Panquesito7 (approved):
LGTM 🚀 Thank you for your contribution! 😄👍

src/Components/Profile.css
```diff
@@ -6,6 +6,8 @@
 }
 
 .p-avatar-image > img {
+  width: 4rem;
```
krishguptadev, line 9: can't this be done with prime-react classes?

nathgoutam93: Avatar is the parent component of the image,
But the image is not exposed , I don't know how to apply custom CSS to a child component from its parent component

krishguptadev (approved):

PR #823, opened 2021-12-17
[feat]: Added react-icons and removed primeicons 

Rupesh-1302:
## Fixes Issue
closes #819 and #757

## Changes proposed

- [X] Having react-icons included amongst the packages installed on running `npm install`.
- [X] Replacement of primereact icons (or integrating those with react-icons, if possible), wherever we're using them.
- [X] Adding support for the social link types that don't have any icons at the moment.

## Check List (Check all the applicable boxes) 

- [X] My code follows the code style of this project.
- [X] My change requires changes to the documentation.
- [ ] I have updated the documentation accordingly.
- [X] All new and existing tests passed.
- [X] This PR does not contain plagiarized content.
- [X] The title of my pull request is a short description of the requested changes.

<a href="https://gitpod.io/#https://github.com/EddieHubCommunity/LinkFree/pull/823"></a>

PR #811, opened 2021-12-14
Fix: Icon names work Capitalised and lowercase

nathgoutam93:
## Fixes Issue

closes #805

## Check List 

- [x] My code follows the code style of this project.
- [ ] My change requires changes to the documentation.
- [ ] I have updated the documentation accordingly.
- [x] All new and existing tests passed.
- [x] This PR does not contain plagiarized content.
- [x] The title of my pull request is a short description of the requested changes.

<a href="https://gitpod.io/#https://github.com/EddieHubCommunity/LinkFree/pull/811"></a>

PR #846, opened 2021-12-22
fix: add Abdus Salam

abdussalam02:
## Check List (Check all the applicable boxes) 

- [x] My code follows the code style of this project.
- [ ] My change requires changes to the documentation.
- [ ] I have updated the documentation accordingly.
- [ ] All new and existing tests passed.
- [x] This PR does not contain plagiarized content.
- [ ] The title of my pull request is a short description of the requested changes.

## Note to reviewers

<a href="https://gitpod.io/#https://github.com/EddieHubCommunity/LinkFree/pull/846"></a>

PR #851, opened 2021-12-23
Update GetIcons.js

krishguptadev:
HotFix 🚨 #823 removed #811 functionality, breaking a lot of files

<a href="https://gitpod.io/#https://github.com/EddieHubCommunity/LinkFree/pull/851"></a>


emmalearnscode:
Did it completely remove the code somehow or is the code still there but now not doing anything?

nathgoutam93:
So, don't you have to remove the changes made with the PR #811 in this PR? 
Or, there is an easy way to do that. 
Just like unmerge  a PR?

emmalearnscode:
Ok, just looking through the files. All the .toLowerCase code is still there in user profile component so I’m not sure why this has broken. Do we still need all the extra code in user profile component if the lower casing can be applied like this directly in getIcon?

nathgoutam93:
> Ok, just looking through the files. All the .toLowerCase code is still there in user profile component so I’m not sure why this has broken. Do we still need all the extra code in user profile component if the lower casing can be applied like this directly in getIcon?

The lower casing can be directly done in get icons component , we can remove the lower casing code from user profile without any issues.

Btw, what kind of broken files you were referring to?
I didn't saw anything broken on live site. 🙄

emmalearnscode:
> The lower casing can be directly done in get icons component , we can remove the lower casing code from user profile without any issues.

I’m not so sure as I think the icon name might also be used for styling, we’ll need to check it to make sure

krishguptadev:
The issue is here afaik https://github.com/EddieHubCommunity/LinkFree/pull/846#issuecomment-999789484

krishguptadev (commented):
Update: The issue was that there was no telegram icon 🤦

emmalearnscode:
Can you add a telegram hover colour to the config/icons.json file also?

Rupesh-1302:
You can also correct arrow left icon it is not lower cased instead It is camel cased in switch case

emmalearnscode (approved):

emmalearnscode:
We now have two identical pull requests for the same issue. As this one is a bit all over the place in the discussion (as it refers to other issues to start with) I'll close this one to make sure the wrong issues don't get closed when the PR is merged. Thanks for contributing :)

PR #863, opened 2021-12-26
chore: add Devesh Jain

devesh2511:
## Check List (Check all the applicable boxes) 

- [x] My code follows the code style of this project.
- [x] My change requires changes to the documentation.
- [x] I have updated the documentation accordingly.
- [x] All new and existing tests passed.
- [x] This PR does not contain plagiarized content.
- [x] The title of my pull request is a short description of the requested changes.

## Note to reviewers

<a href="https://gitpod.io/#https://github.com/EddieHubCommunity/LinkFree/pull/863"></a>

PR #868, opened 2021-12-29
Add/Improve Prettier & Eslint Setup

### LINKED ISSUE #867: add prettier
### Description

We're having some PRs failing the prettier-check, because the prettier-check runs before the actual write. To prevent this we need to take care of prettier in the development phase or adjust the action in my opinion.

Cahllagerfeld:
closes #867 

adds prettier with lint-staged to auto-format all edited files on pre-commit

<a href="https://gitpod.io/#https://github.com/EddieHubCommunity/LinkFree/pull/868"></a>


Cahllagerfeld:
If we decide to update the action instead, feel free to close this.

emmalearnscode (approved):
LGTM 👍

Cahllagerfeld:
Thanks for your feedback Emma :D I'd like to hear some opinions on that before we merge though :D

emmalearnscode:
Sure! I think the discussion came up before about this too and the reason for the action was that not everyone clones to their machine when adding a json file but adds it directly in GitHub and therefore any pre-commit hooks wouldn’t always run but I don’t know enough about it to know

Cahllagerfeld:
Oh thats a good point, I didnt think about this! But then its not very clever to run `prettier --check` before the `prettier --write`. If theres a failure in the check, it wont reacht the actual formatting step

Cahllagerfeld:
Thats right, but take a look at [this](https://github.com/EddieHubCommunity/LinkFree/pull/863) PR. The action fails in the `check` step and doesnt reacht the actual format one

emmalearnscode:
> Thats right, but take a look at [this](https://github.com/EddieHubCommunity/LinkFree/pull/863) PR. The action fails in the `check` step and doesnt reacht the actual format one

Yes, and now there is a third profile that is failing the check too because of formatting. Is it just a case of switching the order things happen in the action or do we need to do something else? I’m not that knowledgeable about actions yet

emmalearnscode:
I notice also the the current action is only looking at json files, I wonder if we should extend that to include other file types too

Cahllagerfeld:
> > Thats right, but take a look at [this](https://github.com/EddieHubCommunity/LinkFree/pull/863) PR. The action fails in the `check` step and doesnt reacht the actual format one
> 
> Yes, and now there is a third profile that is failing the check too because of formatting. Is it just a case of switching the order things happen in the action or do we need to do something else? I’m not that knowledgeable about actions yet

The question is, when switching the steps, if the `check` is still needed. When the check fails after the formatting was successfull, we might have a much bigger issue I think :D

schmelto (approved):

Pradumnasaraf (approved):

Panquesito7 (approved):
Nice work on this! 🚀

Cahllagerfeld:
So I'll merge this if everyone is okay with it 👍

Panquesito7:
> So I'll merge this if everyone is okay with it 👍

Feel free to do it! We got more than 2 approvals from maintainers.

PR #857, opened 2021-12-24
fix: add Telegram icon and its relevant color

BSlaven:
Fix: #853 

I added a fix to the icon and text for telegram in this PR.
I only later realised that there is also an issue raised only for the icon problem.
I added the telegram icon in GetIcons.js file. 
Also added telegram light bluish color to config/links.json file.
Results:
Screenshot (43)

Hover: 
Screenshot (44)

<a href="https://gitpod.io/#https://github.com/EddieHubCommunity/LinkFree/pull/857"></a>

PR #874, opened 2021-12-30
Fix: add default icon color 

### LINKED ISSUE #860: Text not showing when hover
### Description

an issue about the text on the links not showing when hovering with the mouse (see screen shot below)

### Screenshots

Screen Shot 2021-12-26 at 2 58 11 PM

### Additional information

its for the website link the other links work

nathgoutam93:
## Fixes Issue
Closes #860 

## Changes proposed

If icon does not exist in the icons list it is using globe icon as default ,
So, changed the mouse over color to use globe color as well.

## Check List 

## Screenshots

firefox_N4cqVkTsUN
firefox_Y2OZheVOeo

<a href="https://gitpod.io/#https://github.com/EddieHubCommunity/LinkFree/pull/874"></a>


emmalearnscode (changes requested):
This is going to cause problems for all other hover states which are not a globe, what we need to do instead is have a check that link.icon exists and if it doesn’t pass in globe as the default.

Panquesito7 (changes requested):
Hey there! This is not how it should work, I think 😅 You have to do similar steps that are done in this PR https://github.com/EddieHubCommunity/LinkFree/pull/857 but for the `globe` icon. Let us know if you need any help! 😃

nathgoutam93:
> 

I don't think it will cause any problem for other hover states.
As we are filtering links two times:
1.links which have valid icon i.e. icons which are present in the colors list
2.links which have invalid icon i.e. icons which are not present in the colors list

if icon is not in colors list we are already defaulting to globe icon.
So I'm defaulting the color to the globe color as well.

How it will cause problem?

nathgoutam93:
> Hey there! This is not how it should work, I think 😅 You have to do similar steps that are done in this PR #857 but for the `globe` icon. Let us know if you need any help! 😃

I looked into the solution of suggested PR but I think that was a different case, as the Icon was totally not there so he added the icon and also added the color for the icon .

but in this case the icon and the color is already there. (globe)

this PR is not to only fix the website icon issue but also to fix any icons hover effect which are not in the get icon component as well in colors list.

please let me know , if I miss something 😅

emmalearnscode (approved):
Ah yes, I looked at it properly and understand what you did now! LGTM!

nathgoutam93:
> Ah yes, I looked at it properly and understand what you did now! LGTM!

great , thanks for the review . I should have titled the PR properly then it would be better to understand and will make sense what I did.

Vyvy-vi (approved):
LGTM

PR #927, opened 2022-01-07
ci: add lighthouse CI

krshdev:
This is not the final copy of the action probably, this is just to get started, we can improve it as we use :D

@eddiejaoude you need to do this before this PR can be merged: https://github.com/GoogleChrome/lighthouse-ci/blob/main/docs/getting-started.md#github-app-method-recommended

<a href="https://gitpod.io/#https://github.com/EddieHubCommunity/LinkFree/pull/927"></a>

### DIFF
--- a/.github/workflows/lighthouse.yml
+++ b/.github/workflows/lighthouse.yml
@@ -6,10 +6,10 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
-      - name: Use Node.js 12.x
+      - name: Use Node.js 16.x
         uses: actions/setup-node@v1
         with:
-          node-version: 12.x
+          node-version: 16.x
       - name: npm install, build
         run: |
           npm install
--- a/.github/workflows/lighthouse.yml
+++ b/.github/workflows/lighthouse.yml
@@ -1,11 +1,20 @@
-name: Audit live site
-on: [push, pull_request]
-
+name: Lighthouse CI
+on: [push]
 jobs:
-  audit:
+  lhci:
+    name: Lighthouse
     runs-on: ubuntu-latest
     steps:
-    - name: Audit live URL
-      uses: jakejarvis/lighthouse-action@master
-      with:
-        url: 'https://jarv.is/'
+      - uses: actions/checkout@v2
+      - name: Use Node.js 12.x
+        uses: actions/setup-node@v1
+        with:
+          node-version: 12.x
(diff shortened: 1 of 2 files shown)


emmalearnscode:
This repo has the action I mentioned on the livestream. You also need to add a config file though and allow lighthouse ci access to the repo. I’m not sure how the two compare, but mine stops a deploy if scores fall below a certain level. We maybe need to decide if that is something we want or if lighthouse scores are just a nice to have that we can keep an eye on and raise issues ourselves if values fall below a certain level. https://github.com/emmalearnscode/learn-cool-problems/blob/main/lighthouserc.json

emmalearnscode (changes requested):
As Cahl has a PR open to update node version to 16, it’s probably a good idea that this one matches too.

.github/workflows/lighthouse.yml
```diff
@@ -0,0 +1,20 @@
+name: Lighthouse CI
+on: [push]
+jobs:
+  lhci:
+    name: Lighthouse
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - name: Use Node.js 12.x
```
krishguptadev, line 9:
```suggestion
      - name: Use Node.js 16.x
```

.github/workflows/lighthouse.yml
```diff
@@ -0,0 +1,20 @@
...
+        uses: actions/setup-node@v1
+        with:
+          node-version: 12.x
```
krishguptadev, line 12:
```suggestion
          node-version: 16.x
```

krshdev:
Hi @emmalearnscode, could you review this once? 🙂

.github/workflows/lighthouse.yml
```diff
@@ -0,0 +1,20 @@
+name: Lighthouse CI
+on: [push]
+jobs:
+  lhci:
+    name: Lighthouse
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - name: Use Node.js 16.x
+        uses: actions/setup-node@v1
```
nhcarrigan, line 10:
```suggestion
        uses: actions/setup-node@v3
```
v1 is very outdated.

nhcarrigan (changes requested):

krshdev:
#1152 (Sorry, I think I deleted the branch)

PR #937, opened 2022-01-08
feat: Filter by profile type

### LINKED ISSUE #822: Searching by Community or Personal profile
### Description

I think we can improve search function without changing much of the Search bar UI, we can use something like type:community|personal, just like github search functionality __is:open__, etc... .Also, to make the search work either we have to make two seperate folders for containing personal profiles and other with community profile __OR__ we can decide to have a better naming syntax for the json data files like community profiles starting with __comm-username.json__ or any other suggested by the community.

### Screenshots

_No response_

### Additional information

_No response_

nathgoutam93:
## Fixes Issue

Closes #822

## Check List

## Screenshots

https://user-images.githubusercontent.com/91387097/148654527-9f41549c-8628-4a8d-b01c-d5cf4e65a729.mp4

<a href="https://gitpod.io/#https://github.com/EddieHubCommunity/LinkFree/pull/937"></a>

### DIFF
--- a/src/Components/Home/filterProfileType.js
+++ b/src/Components/Home/filterProfileType.js
@@ -5,7 +5,7 @@ export default function ProfileTypeFilter({ profileType, typeHandler }) {
   return (
     <select value={profileType} onChange={(e) => typeHandler(e.target.value)}>
       <option value="personal">Personal</option>
-      <option value="misc">Misc</option>
+      <option value="all">All</option>
       <option value="community">Community</option>
     </select>
   )
--- a/src/Components/Home/filterProfileType.js
+++ b/src/Components/Home/filterProfileType.js
@@ -1,0 +1,24 @@
+import React from 'react'
+import PropTypes from 'prop-types'
+
+export default function ProfileTypeFilter({ profileType, typeHandler }) {
+  return (
+    <div className="flex flex-column justify-content-center">
+      <div className="m-3 flex">
+        <select
+          value={profileType}
+          onChange={(e) => typeHandler(e.target.value)}
+        >
+          <option value="personal">Personal</option>
+          <option value="misc">Misc</option>
+          <option value="community">Community</option>
+        </select>
+      </div>
+    </div>
+  )
+}
+
+ProfileTypeFilter.propTypes = {
+  profileType: PropTypes.string.isRequired,
+  typeHandler: PropTypes.func.isRequired,
+}
--- a/src/Components/Home/filterProfileType.js
+++ b/src/Components/Home/filterProfileType.js
@@ -1,13 +1,20 @@
 import React from 'react'
 import PropTypes from 'prop-types'
+import { Dropdown } from 'primereact/dropdown'
 
 export default function ProfileTypeFilter({ profileType, typeHandler }) {
+  const profileTypes = [
+    { label: 'Personal', value: 'personal' },
+    { label: 'All', value: 'all' },
+    { label: 'Community', value: 'community' },
+  ]
+
   return (
-    <select value={profileType} onChange={(e) => typeHandler(e.target.value)}>
-      <option value="personal">Personal</option>
-      <option value="all">All</option>
-      <option value="community">Community</option>
-    </select>
+    <Dropdown
+      value={profileType}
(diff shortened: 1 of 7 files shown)


emmalearnscode (changes requested):
I think 'all' would be a better word to use instead of 'misc' as you are showing all profiles by default. Misc infers some sort of randomness to it which there isn't.

nathgoutam93:
> I think 'all' would be a better word to use instead of 'misc' as you are showing all profiles by default. Misc infers some sort of randomness to it which there isn't.

Indeed, I wasn't sure what word to use

krishguptadev (commented):
We have got the wording 'personal' and 'community' already, perhaps you can bring those to the use instead of misc?

nathgoutam93:
> We have got the wording 'personal' and 'community' already, perhaps you can bring those to the use instead of misc?

I'm already using them , but as a default showing misc profiles both personal and community . But as @emmalearnscode  suggested I'll change the "misc" with "All"

emmalearnscode (changes requested):
Since we are using prime react, could you change the dropdown to this one so that we keep consistent styles? https://www.primefaces.org/primereact/dropdown/

Thank you :) (And apologies I didn't notice it before)

nathgoutam93:
> Since we are using prime react, could you change the dropdown to this one so that we keep consistent styles? https://www.primefaces.org/primereact/dropdown/
> 
> Thank you :) (And apologies I didn't notice it before)

yup I tried but the problem is that , as primereact had the dependency of prime icons the icons for some components stopeed showing as they do not exist anymore after removal of primeicons . I also tried to use react icons in the component but when i look into the source of drop down component it do not take icons directly instead it takes css class. I'll try to come up with a solution if not then maybe we will need to add prime icons back as it is a dependency of preimereact

emmalearnscode:
> as primereact had the dependency of prime icons

If the components depend on prime icons then they should be installed back into the project, otherwise we will run into the same issue every time we try to use a prime react component. It shouldn't be a problem to have both icon packages there as far as I know.

nathgoutam93:
> > as primereact had the dependency of prime icons
> 
> If the components depend on prime icons then they should be installed back into the project, otherwise we will run into the same issue every time we try to use a prime react component. It shouldn't be a problem to have both icon packages there as far as I know.

should I install primeicons in this PR or an issue should be raised?

nathgoutam93:
firefox_xhpFiYKTAX

as you can see in screenshot the drop down icon is not working

emmalearnscode:
Add it with this PR as it is something you need to get the dropdown working properly

emmalearnscode (approved):
Looks like it's working well to me. Let's just wait for a couple more reviews to make sure nothing has been missed :)

Panquesito7 (approved):
LGTM 🚀 Thank you for your contribution! 😄👍

src/Components/Home/filterProfileType.js
```diff
@@ -0,0 +1,24 @@
+import React from 'react'
+import PropTypes from 'prop-types'
+import { Dropdown } from 'primereact/dropdown'
+
+export default function ProfileTypeFilter({ profileType, typeHandler }) {
+  const profileTypes = [
+    { label: 'Personal', value: 'personal' },
+    { label: 'All', value: 'all' },
+    { label: 'Community', value: 'community' },
```
schmelto, line 9: I would prefer to make the selection in alphabetical order. 

image

```suggestion
    { label: 'All', value: 'all' },
    { label: 'Community', value: 'community' },
    { label: 'Personal', value: 'personal' }
```

schmelto (changes requested):

src/Components/Home/filterProfileType.js
nathgoutam93: firefox_h8Ha0XXI8K

I think length order will look more neat ,what you say?

schmelto: I would stick to alphabetical but I leave it for discussion.
But I think we can convert this to a new issue and merge your changes.

schmelto (approved):

PR #992, opened 2022-01-14
Add tabIndex attribute to scroll to top btn

### LINKED ISSUE #990: To Top button not accessible with keyboard
### Description

The return to top button is currently not accessible to anyone using a keyboard to tab through the site. The button needs a tabindex=0 attribute adding and once pressed the focus of the site should be moved to the top of the page.

### Screenshots

_No response_

### Additional information

_No response_

BSlaven:
Closes #990 

Simply adding tabIndex attribute to the button did not work in the sense that pressing Enter did not trigger the function.

My solution was to add onKeyPress attribute to the button and pass it the same function as for onClick.

This of course meant that I had to make sure that only pressing 'Enter' and clicking on the button trigger that function.
Hence the condition.

<a href="https://gitpod.io/#https://github.com/EddieHubCommunity/LinkFree/pull/992"></a>

### DIFF
--- a/src/Components/ScrollToTopBtn.js
+++ b/src/Components/ScrollToTopBtn.js
@@ -26,7 +26,6 @@ function ScrollToTopBtn() {
         top: 0,
         behavior: 'smooth',
       })
-      setScrollBtnVisibility(scrollBtnVisibility)
     }
   }
 
--- a/src/Components/ScrollToTopBtn.js
+++ b/src/Components/ScrollToTopBtn.js
@@ -20,16 +20,20 @@ function ScrollToTopBtn() {
     }
   }, [])
 
-  const scrollToTop = () => {
-    window.scrollTo({
-      top: 0,
-      behavior: 'smooth',
-    })
-    setScrollBtnVisibility(scrollBtnVisibility)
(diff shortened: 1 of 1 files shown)


emmalearnscode (approved):
Ah yes, that makes sense! Thanks for spotting the need for a key press and adding it in. LGTM!

syedsohan (approved):

digambar-t7:
Hey, can you remove line no 29 and check if it is significant

BSlaven:
Actually line 29 is rather useless. With or without it functionality seems to work fine both on homepage and on profile page.
Also both by pressing enter or clicking on it.
Should I remove it? Thoughts?

digambar-t7:
Being the creator of the scroll-to-top-btn, I think that line no.29 should be erased because I don't remember the grounds on which I included it. 
But, let's have a confirmation from others.

emmalearnscode:
Yes, go ahead and remove it if it’s not serving any purpose

BSlaven:
I removed the unnecessary line.

nathgoutam93 (approved):

PR #1124, opened 2022-02-12
feat: Functionality to cache and reuse dependencies before build

### LINKED ISSUE #1118: [FEATURE] Enable caching of dependencies to save build minutes
### Description

- In `.github/workflows/build.yaml` we can cache the dependencies based on changes in `package-lock.json` .
- This will help make our build much faster.
- Since for free tier, only 2000 build minutes/month are allotted on github actions servers, if in future the project becomes very large, we can take advantage of caching and keep the build times low.

#### I would like to work on this issue
- I will create hashes for the package-lock and setup lookups based on hashes to help fetch the cache if the dependencies are same as before.

**Please assign me this issue if found valid** 🙂 

### Screenshots

_No response_

### Additional information

- Also I found that there was a lot of code duplication among  `.github/workflows/build.yaml` file and the other workflows as well such as checking out to the repo in the VM and then installing all dependencies. 
- I do know a couple ways that might help but not sure about them. I can raise new issue for this and research on it as well or we can discuss with other community members.
- A couple ways that I have for this are making a single reusable template(composite actions) or combining jobs just have separate steps but that would compromise parallelism but then we don't have to setup a new environment again.
- Let me know if this sounds valid, I'll raise a new issue for it then and research on it a bit more.

Pranav016:
## Fixes Issue
Closes #1118 

## Changes proposed
- Add functionality to cache dependencies and reuse the cache if the dependencies haven't change.
- This makes the build process much faster (we can build within seconds)
- Since for free tier, only 2000 build minutes/month are allotted on GitHub action servers, if in future the project becomes very large, we can take advantage of caching and keep the build times low.

<a href="https://gitpod.io/#https://github.com/EddieHubCommunity/LinkFree/pull/1124"></a>

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -9,6 +9,13 @@ jobs:
       - uses: actions/setup-node@v2
         with:
           node-version: '16'
+      - name: cache lookup
+        uses: actions/cache@v2
+        with:
+          path: ~/.npm
+          key: ${{ runner.os }}-node-${{ hashFiles('**/package-lock.json') }}
+          restore-keys: |
+            ${{ runner.os }}-node-
       - name: install dependencies
         run: npm ci
       - name: run linter
@@ -23,6 +30,13 @@ jobs:
       - uses: actions/setup-node@v2
         with:
           node-version: '16'
+      - name: cache lookup
+        uses: actions/cache@v2
+        with:
+          path: ~/.npm
+          key: ${{ runner.os }}-node-${{ hashFiles('**/package-lock.json') }}
+          restore-keys: |
+            ${{ runner.os }}-node-
       - name: install dependencies
         run: npm ci
       - name: run tests
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -10,6 +10,7 @@ jobs:
         with:
           node-version: '16'
       - name: cache lookup
+        id: cache
         uses: actions/cache@v2
         with:
           path: ~/.npm
@@ -17,6 +18,7 @@ jobs:
           restore-keys: |
             ${{ runner.os }}-node-
       - name: install dependencies
+        if: steps.cache.outputs.cache-hit != 'true'
         run: npm ci
       - name: run linter
         run: npm run lint
@@ -31,6 +33,7 @@ jobs:
         with:
           node-version: '16'
       - name: cache lookup
+        id: cache
         uses: actions/cache@v2
         with:
           path: ~/.npm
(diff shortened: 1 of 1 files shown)


Pranav016:
@eddiejaoude can you pls review this PR? 🙂

.github/workflows/build.yml
```diff
@@ -23,6 +30,13 @@ jobs:
       - uses: actions/setup-node@v2
         with:
           node-version: '16'
+      - name: cache lookup
+        uses: actions/cache@v2
```
nhcarrigan, line 34:
```suggestion
      - name: cache lookup
        id: cache
        uses: actions/cache@v2
```

.github/workflows/build.yml
```diff
@@ -9,6 +9,13 @@ jobs:
...
+        with:
+          path: ~/.npm
+          key: ${{ runner.os }}-node-${{ hashFiles('**/package-lock.json') }}
+          restore-keys: |
+            ${{ runner.os }}-node-
       - name: install dependencies
         run: npm ci
```
nhcarrigan, line 20:
```suggestion
      - name: install dependencies
        if: steps.cache.outputs.cache-hit != 'true'
        run: npm ci
```

.github/workflows/build.yml
```diff
@@ -23,6 +30,13 @@ jobs:
...
```
nhcarrigan, line 41:
```suggestion
      - name: install dependencies
        if: steps.cache.outputs.cache-hit != 'true'
        run: npm ci
```

nhcarrigan (changes requested):
My understanding is we actually need to check if the cache step has a hit, and skip the install step if it does.

krshdev (commented):
I looked up the documentation for this, and I had a question, did you have a look at these? 

image

Pranav016:
> I looked up the documentation for this, and I had a question, did you have a look at these?
> 
> image

It only breaks if we are never updating the cache(which usually happens with our local cache) but in this case we are updating the cache whenever our lock file is changed and we store cache even for different node versions if we decide to test on multiple. I have cached my dependencies in the past, even with `npm ci` and they haven't given any issues..
Can you also provide a link to the reference docs, I couldn't find the dedicated note.

krshdev:
> Can you also provide a link to the reference docs, I couldn't find the dedicated note.

https://github.com/actions/cache/blob/main/examples.md#node---npm

nhcarrigan:
That's a good catch @krshdev! It sounds like this isn't an approach we want to take, then.

nhcarrigan:
Going to close this per Eddie's comment above. Thanks for your interest in contributing!!!

PR #1171, opened 2022-03-08
feat: add text on main page

### LINKED ISSUE #1123: [FEATURE] Add text on main page
### Description

The reason I created #1077 was because I had a idea, the icon can show up here with the text "Linkfree" in a epic font:

image

I only think this will be good on the MAIN/HOME page. 

### Screenshots

_No response_

### Additional information

If, there's coding needed in HTML for this, I can help, if in Node, won't be possible.

santhoshbala0178:
## Changes proposed
- Added "LinkFree" text along with Icon in Home Page and User Page
- Added Loading state in Navbar.js to prevent items moving around
- Added images folder under "src" to acces the logo

## Screenshots
UserPage
HomePage

## Note to reviewers
- Added a new folder "images" under "src" folder to add the icon, as because of an existing issue (https://github.com/facebook/create-react-app/issues/9937) we will not be able to access images directly from public folder

<a href="https://gitpod.io/#https://github.com/EddieHubCommunity/LinkFree/pull/1171"></a>


emmalearnscode (changes requested):
Thanks for your contribution. Here's a few things I noticed and it would be great if you could work on them:
- the link doesn't work
- the icon is too small
- on small screens the menubar layout needs fixing 
- the image and text are not removed from the DOM when the screen gets smaller
- The image/text is not centered on the profile page

src/Components/UserProfile/User.js
```diff
@@ -30,7 +30,13 @@ function User() {
     <>
       <Navbar
         start={
-          <Link to="/" aria-label="Go back to Home">
+          <Link
+            to="/"
+            aria-label="Go back to Home"
+            style={{
+              'margin-right': '15rem', // Added to center the icon & label
```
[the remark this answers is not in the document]
santhoshbala0178: Modified as suggested.

src/Components/Navbar.css
```diff
@@ -1,3 +1,32 @@
...
+  flex-wrap: wrap;
+  justify-content: space-between;
+}
+
+.p-menubar-end {
+  margin-left: 8rem !important;
+}
+
+.p-menuitem-link {
+  cursor: default !important;
```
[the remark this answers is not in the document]
santhoshbala0178: I have assumed that the text would just be displayed without any navigation, Resolved now with redirection to Home Page.

santhoshbala0178:
@emmalearnscode, Thanks for the feedback. I have resolved all of the mentioned issues except one,

 - the image and text are not removed from the DOM when the screen gets smaller

This is actually handled by the _primereact_ library to hide the menu items below certain width, should I work on overriding all of the properties.?

```
@media screen and (max-width: 960px)
.p-menubar .p-menubar-root-list {
    position: absolute;
    display: none;
    padding: 0.25rem 0;
    background: #ffffff;
    border: 0 none;
    box-shadow: 0 2px 4px -1px rgb(0 0 0 / 20%), 0 4px 5px 0 rgb(0 0 0 / 14%), 0 1px 10px 0 rgb(0 0 0 / 12%);
    width: 100%;
}
```

src/Components/Home/Searchbar.js
```diff
@@ -15,6 +15,9 @@ const Searchbar = ({ searchHandler, searchTerm }) => {
           name="user"
           id="search-input"
           placeholder="Search user..."
+          style={{
```
emmalearnscode, line 18: I'm not sure that this style makes any difference. I've tried it without and can't see that it does anything.

src/Components/Navbar.css
```diff
@@ -1,3 +1,41 @@
...
+  background-image: url('../images/linkfree.ico');
+  width: 2rem;
+  height: 2rem;
+  background-size: 2rem 2rem;
+}
+
+.p-menuitem-text {
+  font-weight: 600;
+  font-size: 20px;
+}
```
emmalearnscode, line 41: I think it might look better if the icon and title disappeared on screens smaller than 864px, that's where the chips jump to 2 columns instead of 3.

src/Components/UserProfile/User.js
```diff
@@ -30,7 +30,13 @@ function User() {
     <>
       <Navbar
         start={
-          <Link to="/" aria-label="Go back to Home">
+          <Link
+            to="/"
+            aria-label="Go back to Home"
+            style={{
```
emmalearnscode, line 36: This should not be needed in order to center the icon and title. I think there may be some random margin on the github icon/version number that shouldn't be there which is making these things flex wrongly.

emmalearnscode:
> @emmalearnscode, Thanks for the feedback. I have resolved all of the mentioned issues except one,
> 
> * the image and text are not removed from the DOM when the screen gets smaller
> 
> This is actually handled by the _primereact_ library to hide the menu items below certain width, should I work on overriding all of the properties to show the text in smaller screens as well.?
> 
> ```
> @media screen and (max-width: 960px)
> .p-menubar .p-menubar-root-list {
>     position: absolute;
>     display: none;
>     padding: 0.25rem 0;
>     background: #ffffff;
>     border: 0 none;
>     box-shadow: 0 2px 4px -1px rgb(0 0 0 / 20%), 0 4px 5px 0 rgb(0 0 0 / 14%), 0 1px 10px 0 rgb(0 0 0 / 12%);
>     width: 100%;
> }
> ```

Don't worry about it just now. Unfortunately the prime react library isn't the greatest when it comes to accessibility.

src/Components/Home/Searchbar.js
santhoshbala0178: Removed the style.

src/Components/Navbar.css
santhoshbala0178: As mentioned in the comment below, this is handled by the primereact library to hide for screens smaller than 960px. We may need to override many properties to achieve this.

src/Components/UserProfile/User.js
santhoshbala0178: If the margin is not added to the github icon and the content is centered only using justify-content: space-between, the icon and title are not centered to the page properly.

AlphaVS-76 (approved):

krshdev (approved):

emmalearnscode:
Thanks for your contribution. In the end we decided to take the home page in a different direction which has led to some merge conflicts in the pull request. If you want to have a go at fixing them then that would be great but we also understand if you would rather close this pull request in case it is too much work and you don't have the time.

PR #1178, opened 2022-03-13
Create sheoranharis.json

sheoranharis:
Created LinkFree website

<a href="https://gitpod.io/#https://github.com/EddieHubCommunity/LinkFree/pull/1178"></a>

PR #1184, opened 2022-03-17
Revert "Create sheoranharis.json"

kabir0x23:
Reverts EddieHubCommunity/LinkFree#1178

having issues with files

users file must be in the public dir but in this case, it is in the root dir
and the user is not accessible either from the website
image

<a href="https://gitpod.io/#https://github.com/EddieHubCommunity/LinkFree/pull/1184"></a>


Pradumnasaraf (changes requested):
Instead of reverting this PR, we can change the file path.

In this way, the profile will also not be removed, and don't have to create another PR for adding the profile.

PR #1435, opened 2022-07-03
Add icons usage terms

### LINKED ISSUE #845: [DOCS] update icons table with icon name to use in components/json
### Description

The table of icons currently has the name and icon, however the name doesn’t always correspond to what is used in components/json. For example, PayPal is paypal. graduation hat is graduation-hat etc. I suggest either adding an extra column for usage instructions to the table or including it alongside the name in the name box.

### Screenshots

5D8B5FE5-2D97-4BA1-A374-3E06BFC194F9

### Additional information

_No response_

BSlaven:
Closes #845 

## Changes proposed
I added usage terms in icons.md file. These terms correspond to terms found in links.json file.

## Note to reviewers
Some icons do not exist in links.json file (e.g. search and left arrow).
This is a potential different issue.
I left them empty as can be seen in screenshot below:
Screenshot (3)

On the other hand hackerrank does not have an icon in icons.md.
I simply ignored it.
This is also a potential issue yet to be created.


kumarsonsoff3 (commented):
I don't think we need this, because as per my knowledge, these icons usage is not case specific. 

Like `Hashnode` and `hashnode` both works fine.

kumarsonsoff3:
I've checked a few more profiles, it just works fine. Icon usage is not case specific. 
Even `YouTube`, `Youtube`, `youtube`, all works fine!

emmalearnscode:
> I don't think we need this, because as per my knowledge, these icons usage is not case specific.
> 
> Like `Hashnode` and `hashnode` both works fine.

This is something that was requested a long time ago as it's easy to miss that graduation-hat needs a hyphen. I think this is a great way to make our instructions even clearer.

emmalearnscode (changes requested):
This looks great but there appears to be a merge conflict. If you could take a look at that then hopefully we can get this file updated.

PR #1572, opened 2022-08-05
[chore]: fixed class in css

kartikgajjar7:


EmmaDawsonDev (commented):
These are not referring to classes but to elements and therefore this change is not required.

PR #1776, opened 2022-09-20
ci: release + deployment on cron

eddiejaoude:
Create Release and then trigger Deploy every night at midnight UTC

Manual tiger moved to Release

No longer creates release for every push to `main`, will collect the days work up together and changelog will have more changes in it

<a href="https://gitpod.io/#https://github.com/EddieHubCommunity/LinkFree/pull/1776"></a>

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -2,7 +2,6 @@ name: Deploy
 on:
   release:
     types: [published]
-  workflow_dispatch:
 
 jobs:
   build:
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,5 +1,6 @@
 name: Releases
 on:
+  workflow_dispatch:
   schedule:
     - cron:  '00 00 * * *'
 
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,8 +1,7 @@
 name: Releases
 on:
-  push:
-    branches:
-      - main
+  schedule:
+    - cron:  '00 00 * * *'
 
 jobs:
   changelog:
--- a/README.md
+++ b/README.md
@@ -7,7 +7,7 @@
 
 > **Warning**
 >
-> Your profile will need to be deployed to appear, we try to collect a few together before deploying
+> Your profile will need to be deployed to appear, deployments happen at midnight UTC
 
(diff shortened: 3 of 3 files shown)


mocdaniel:
Do we need to add a remark to the `README.md` that all changes merged across a day's range will go live at 00:00 UTC?

.github/workflows/deploy.yml
```diff
@@ -2,7 +2,6 @@ name: Deploy
 on:
   release:
     types: [published]
-  workflow_dispatch:
```
naomi-lgbt, line 5:
```suggestion
  workflow_dispatch
```

We should leave this IMHO. For times we want to run a manual deployment.

.github/workflows/release.yml
```diff
@@ -1,8 +1,8 @@
 name: Releases
 on:
-  push:
-    branches:
-      - main
+  workflow_dispatch:
+  schedule:
+    - cron:  '00 00 * * *'
```
naomi-lgbt, line 6: I don't believe this is the action we want to CRON.

The problem with this action is if all we have in the change is `chore` commits (which is accurate for profile updates), the changelog doesn't show a version bump so nothing is actually triggered. Refer to [this log](https://github.com/EddieHubCommunity/LinkFree/actions/runs/3042239940/jobs/4900198039) for an example.

The CRON should be on the dispatch workflow instead.

naomi-lgbt (changes requested):

.github/workflows/deploy.yml
eddiejaoude: I think it should always be run from the `Release` Action though, so I moved it from `deploy` to `release` Action - that way a release is not created on every commit/push but manually or on the cron (the changelogs will be much more interesting too)

.github/workflows/release.yml
eddiejaoude: Good point however the deploy only deploys releases, so that would not work either 🤔

naomi-lgbt: The deploy was the script we were running to trigger a manual deployment. We cannot CRON the release action because it will exit if it doesn't see a version-changing commit.

eddiejaoude:
Going to close this for now

PR #1845, opened 2022-10-01
perf: compress site

krshkun:
I think it will be a good idea to compress the site and also serve images in webp format, as per my understanding if nextjs converts them to webp, we won't face the api errors

<a href="https://gitpod.io/#https://github.com/EddieHubCommunity/LinkFree/pull/1845"></a>


krshkodes:
Also unsure why tests ran before build...

eddiejaoude:
Thanks Krish! But not sure if `webp` is supported on Mac yet though

krshkun:
You would be surprised to know, yes, they decided to support webp but not svg: https://caniuse.com/webp

eddiejaoude:
Thanks Krish 👍 It would be great to know the size comparison before and after?

Btw I am curious why these are not on by default? There must be a reason?

krshkun:
NextJS wouldn't fiddle with the quality of the images - but because the avatars/images we don't use aren't so detailed already - we don't need that...

Plus, most websites use it including Discord as well :D 

I am not sure about comparison - I'll have to install a bunch more code so that it checks the bundle size each time - but that sounds like not having a place in this PR,

but:

> WebP lossless images are 26% smaller in size compared to PNGs.
-- developers.google.com

eddiejaoude (approved):
Thanks Krish 👍

Issue #1927, opened 2022-10-08
Dark mode toggle is not reachable with keyboard navigation


EmmaDawsonDev:
### Description

<img width="74" alt="image" src="https://user-images.githubusercontent.com/57045550/194723734-c74a9280-e4cd-4c3b-a27e-98f058311116.png">

The toggle button in the header is unreachable with the tab key when navigating by keyboard. It should be possible to reach it and activate it just by tabbing through the site.

vasucp1207:
Hey, @EmmaDawsonDev I want to work on this issue

vasucp1207:
Hey, @EmmaDawsonDev PR done take a look at that.

eddiejaoude:
The new version of LinkFree is out, I will close this issue, feel free to create a new issue for the new version if still relevant

PR #1931, opened 2022-10-08
Fixing Dark mode toggle is not reachable with keyboard navigation

vasucp1207:
Fix Issue #1927 
Screenshot 2022-10-09 at 3 36 07 AM

<a href="https://gitpod.io/#https://github.com/EddieHubCommunity/LinkFree/pull/1931"></a>

### DIFF
--- a/src/Components/Navbar.js
+++ b/src/Components/Navbar.js
@@ -43,7 +43,7 @@ function Navbar({ items, start, end }) {
 
         <div style={theme}>v{version}</div>
 
-        <div className="theme--button">
+        <a href="#" className="theme--button">
           {darkTheme
             ? (
             <GetIcons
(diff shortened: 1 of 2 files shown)


src/Components/Navbar.js
```diff
@@ -43,7 +43,7 @@ function Navbar({ items, start, end }) {
 
         <div style={theme}>v{version}</div>
 
-        <div className="theme--button">
+        <a href="#" className="theme--button">
```
EmmaDawsonDev, line 46: Link is not the right semantic element to choose here because it's not leading anywhere. A button would be the most appropriate as it performs an action.

EmmaDawsonDev (changes requested):
Please use a button instead.

src/Components/Navbar.js
vasucp1207: Ok I will fix it soon.

eddiejaoude (approved):
Thank you 👍

PR #1966, opened 2022-10-14
Update Gitpod.yml for Next.js

Cahllagerfeld:
Adjustments for Gitpod.yml in the new project setup

### DIFF
--- a/.gitpod.yml
+++ b/.gitpod.yml
@@ -1,3 +1,4 @@
+---
 # List the start up tasks. Learn more https://www.gitpod.io/docs/config-start-tasks/
 tasks:
   - name: Install npm depdencies
@@ -5,9 +6,9 @@ tasks:
     command: |
       gp await-port 27017
       npm run dev
-  - name: Docker
+  - name: Mongo
     command: |
-      docker-compose up
+      mkdir -p /workspace/data && mongod --dbpath /workspace/data
 
 github:
   prebuilds:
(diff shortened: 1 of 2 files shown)


.gitpod.yml
```diff
@@ -1,13 +1,15 @@
+---
 # List the start up tasks. Learn more https://www.gitpod.io/docs/config-start-tasks/
 tasks:
   - name: Install npm depdencies
     init: npm i
     command: |
-      gp await-port 27017
+      mv .env.example .env
```
Cahllagerfeld, line 7: I'm renaming the example env here to a regular one, so the `dev` command executes as expected.

When not doing this, it probably doesn't make sense to run `dev` because the database won't connect.
I personally would go with the solution that is currently in that PR, what do you think?

.gitpod.yml
```diff
@@ -19,9 +21,8 @@ github:
...
     addBadge: true
 
-image:
-  file: .gitpod.dockerfile
+image: gitpod/workspace-mongodb
 
 vscode:
   extensions:
-    - "asciidoctor.asciidoctor-vscode"
+    - bradlc.vscode-tailwindcss
```
Cahllagerfeld, line 28: I also removed the `asciidoctor` extension, because we don't use this anymore.
Other than that I added the tailwind extension, as we bring this in with nextjs, so people have intellisense for the classes.

krshkun (approved):
Looks cool!

.gitpod.yml
eddiejaoude: Sounds good 👍

PR #2101, opened 2022-11-03
fix : updating links for new username

kartikgajjar7:
username is changed so links should be update


kumarsonsoff3 (approved):
Looks Good, Thanks! 🎉 

---
✔️ Verified that new links are of the same person! 
**Reference for confirmation:** Abhishek's other links in file!

eddiejaoude:
PR owner does not match the filename (username), so I will close

PR #2193, opened 2022-11-09
Add MongoDB .env error handler

mkubdev:
## Changes proposed

Added a check in `config/mongo.js` to check if the `LINKFREE_MONGO_CONNECTION_STRING` mongo connection string is present inside `env.local` :

```js
// config/mongo.js
if (!process.env.LINKFREE_MONGO_CONNECTION_STRING) {
  throw new Error(
    "Please define LINKFREE_MONGO_CONNECTION_STRING to .env.local"
  );
}
```

## Note to reviewers

Hello 😁

<a href="https://gitpod.io/#https://github.com/EddieHubCommunity/LinkFree/pull/2193"></a>

### DIFF
--- a/config/mongo.js
+++ b/config/mongo.js
@@ -1,5 +1,11 @@
-import mongoose from "mongoose";
-import * as fs from "fs";
+import mongoose from 'mongoose';
+import * as fs from 'fs';
+
+if (!process.env.LINKFREE_MONGO_CONNECTION_STRING) {
+  throw new Error(
+    'Please define LINKFREE_MONGO_CONNECTION_STRING to .env.local'
+  );
+}
 
 let hasConnection = false;
 const connectMongo = async () => {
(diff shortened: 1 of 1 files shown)


mkubdev:
> Thank you 👍 Maybe the message would be better to say `...environment variable` as it could be deployed somewhere else

Oops, sure! Something like this? 
```js
 throw new Error(
       "Please define LINKFREE_MONGO_CONNECTION_STRING environment variable inside .env file"
  );
  ```

**Comment by eddiejaoude** [comment]
Thanks @mkubdev 

I am still not sure if we should mention `.env` because when deploying we don't use that, maybe

```
"Please define the LINKFREE_MONGO_CONNECTION_STRING environment variable (if local add to .env file)"
```

mkubdev:
> Thanks @mkubdev
> 
> I am still not sure if we should mention `.env` because when deploying we don't use that, maybe
> 
> ```
> "Please define the LINKFREE_MONGO_CONNECTION_STRING environment variable (if local add to .env file)"
> ```

Sounds good! Updating..

eddiejaoude (approved):
Looks good 👍  thanks

PR #2196, opened 2022-11-10
fix: refactor search page structure

schmelto:
**Updated:**

image

user not found under the search bar.

**Old:**
image

This change prevents that the search bar is jumping on the page when the error message appears

Further added a warning message if the search string is less than 3 characters
image

Remove the results list if the search string is less than 3 characters

### DIFF
--- a/pages/search.js
+++ b/pages/search.js
@@ -61,11 +61,6 @@ export default function Search({ users }) {
         <link rel="icon" href="/favicon.ico" />
       </Head>
       <div className="flex flex-col px-6 align-center">
-        {notFound && (
-          <h2 className="bg-red-200 text-red-600 border-2 border-red-600 p-5 my-5 text-xl">
-            {notFound} not found
-          </h2>
-        )}
         <h1 className="text-4xl mb-4  font-bold">Search</h1>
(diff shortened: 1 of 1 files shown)


pages/search.js
```diff
@@ -61,18 +68,23 @@ export default function Search({ users }) {
...
+        {notFound && (
+          <h2 className="bg-red-200 text-red-600 border-2 border-red-600 p-5 my-5 text-xl">
+            {notFound} not found
+          </h2>
+        )}
+        {!threeOrMore && (
+          <h2 className="bg-yellow-200 text-yellow-600 border-2 border-yellow-600 p-5 my-5 text-xl">
+            You have to enter at least 3 characters to search for a user.
+          </h2>
+        )}
```
schmelto, line 87: This is something we should discuss if this is more suitable as the text in the search box

eddiejaoude: we could also move the alert to a separate component in another PR, maybe it accepts a type of alert (info, warning, error)

eddiejaoude (approved):
Looks great 👍 thanks Tom!

Issue #2213, opened 2022-11-12
[FEATURE] Timeline for latest version


vasucp1207:
### Description

- Improve user timeline for the user version.

### Screenshots

https://user-images.githubusercontent.com/85363195/201461094-88e303e3-ec59-4286-803b-89deb7d0ed84.mov

#### @eddiejaoude assign this to me the implementation is almost complete.

eddiejaoude:
It looks good, I think this takes a lot of vertical space, and we have more information coming. We should probably discuss it, maybe if you have started the changes, can you create a draft PR so it can be discussed

eddiejaoude:
This has been refactored with Tailwind UI components

PR #2218, opened 2022-11-12
Timeline update for user Milestones

vasucp1207:
## Fixes Issue https://github.com/EddieHubCommunity/LinkFree/issues/2213

<a href="https://gitpod.io/#https://github.com/EddieHubCommunity/LinkFree/pull/2218"></a>


components/user/UserMilestone.js
```diff
@@ -1,22 +1,50 @@
 import Icon from "../Icon";
+import { VerticalTimelineElement } from 'react-vertical-timeline-component'
+import 'react-vertical-timeline-component/style.min.css'
+import Image from "next/image";
 
 export default function UserMilestone({ milestone }) {
+  function getStyle() {
+    if (milestone.icon === 'github') return 'black'
+    else return milestone.color
```
eddiejaoude, line 9: This doesn't follow the projects coding style

I think either use a ternary or write the if/else out in full

eddiejaoude: As it is only used once in the code, it could be done inline with a ternary

components/user/UserMilestone.js
```diff
@@ -1,22 +1,50 @@
...
+  }
+
+  function getStyleColor() {
+    if (milestone.icon === 'github') return 'white'
+    else return 'white'
```
eddiejaoude, line 14: Same as above

styles/globals.css
```diff
@@ -15,6 +15,21 @@ a {
...
+.vertical-timeline-element-content {
+  box-shadow: rgba(0, 0, 0, 0.35) 0px 5px 15px !important;
+}
+
+@media only screen and (max-width: 1300px) {
+  .vertical-timeline-element-date {
+    display: block !important;
+    float: none !important;
+  }
+}
```
eddiejaoude, line 31: these should probably only be used in the relevant components rather than globally

vasucp1207: @eddiejaoude, this is for the styling of pseudo before selector which is present by default in the timeline component so styling within the element is not possible as they are hidden.

vasucp1207: @eddiejaoude, the date is given as a property in the timeline component so styling date is not possible in tailwind.

eddiejaoude: I think the property can be styled in the timeline component?

vasucp1207: Hey @eddiejaoude, I try some possible cases like if I apply a border to the **VerticleTimelineElement** component then it applies to the entire section and looks like this image, 

Screenshot 2022-11-20 at 10 33 15 PM

Also if I wrap the component items inside a div and apply a border to that div then it looks something like this,

Screenshot 2022-11-20 at 10 38 43 PM

The **date** is set as a property of the timeline so it is not possible to style its container within the component, same is for the pseudo **before** element.

eddiejaoude:
Thank you for your pull request @vasucp1207 👍  . We purchased Tailwindcss UI components to use on the project, and this section has been refactored multiple times already. Maybe you can make further improvements to what we have now using tailwind

Screenshot 2022-12-08 at 08 27 41

Thank you for your contribution, but I will close this to avoid confusion. Any questions let us know

PR #2269, opened 2022-11-24
fix: load json file directly

eddiejaoude:
Improvements the performance of the profile page

- no need to filter files
- we know the filename


mkubdev (approved):
Freshly tested these changes using `localhost:3000/api/users/mkubdev`, and everything looks good to me. during the JSON parsing.

Issue #2288, opened 2022-11-28
Standardize padding on main pages


Balastrong:
### Description

I noticed that `padding-top` and `padding-left` on the main pages (`/popular`, `/search` and `/events`) are not consistent, giving a weird effect when switching between them.

The popular page has the class `.p-5` on the first element, giving 1.25 rem on both top and left paddings.
Instead, search and events have on their first elements the class `.px-6` giving 1.5 rem on left and 0 on top.

I think they should be standardized (I would suggest to stick to `.p-5` on all three pages)

### Screenshots

![popular](https://user-images.githubusercontent.com/7253929/204372138-a43ef0d4-0918-4bbf-95cf-5f81d6c3e59a.png)

![search](https://user-images.githubusercontent.com/7253929/204372166-1eb8ab1b-62fc-4d21-a4d1-451ada281411.png)

![events](https://user-images.githubusercontent.com/7253929/204372189-bb21ff6a-0693-4c80-ab15-20958f16a377.png)

Balastrong:
If you think it makes sense, I'll be happy to submit a PR for this :)

AAYUSH-droid:
Would like to work on this issue. Can you assign me?

schmelto:
@Balastrong  I assigned you
@AAYUSH-droid  maybe you can do a pair with @Balastrong

Balastrong:
Fixed with #2336

PR #2336, opened 2022-12-07
feat: create Page component to unify padding and sizing of the main pages

Balastrong:
## Changes proposed

As described in #2288 there is some inconsistency between the main pages.
This PR adds a `Page` component that can be used on the main pages to make sure the styling is consistent.

## Screenshots

popular

search

events

<a href="https://gitpod.io/#https://github.com/EddieHubCommunity/LinkFree/pull/2336"></a>

### DIFF
--- a/components/Page.js
+++ b/components/Page.js
@@ -1,0 +1,7 @@
+export default function Page({ children }) {
+  return (
+    <div className="max-w-7xl mx-auto flex flex-col p-6 align-center">
+      {children}
+    </div>
+  );
+}
--- a/pages/[username].js
+++ b/pages/[username].js
@@ -10,6 +10,7 @@ import UserLinks from "../components/user/UserLinks";
 import UserMilestones from "../components/user/UserMilestones";
 import UserTestimonials from "../components/user/UserTestimonials";
 import UserEvents from "../components/user/UserEvents";
+import Page from "../components/Page";
 
 export async function getServerSideProps(context) {
   let data = {};
@@ -98,7 +99,7 @@ export default function User({ users, data, BASE_URL }) {
         <meta property="og:image" content={data.avatar} />
       </Head>
 
-      <div className="mx-auto container px-6 mt-6">
+      <Page>
         <UserProfile data={data} BASE_URL={BASE_URL} />
 
         <UserTabs tabs={tabs} setTabs={setTabs} />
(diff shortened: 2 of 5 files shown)


components/Page.js
```diff
@@ -0,0 +1,7 @@
+export default function Page({ children }) {
+  return (
+    <div className="max-w-7xl mx-auto flex flex-col p-6 align-center">
+      {children}
+    </div>
+  );
+}
```
eddiejaoude, line 7: Would this be better to be added to the layout instead of a new component?

Balastrong: The idea of a new component was to make sure all main pages remain consistent.

I think I did not understand where you would put it instead. In the two files in `components/layouts`?

Balastrong: Ah ok, the problem I see there is that in that case it would also affect the homepage, adding unwanted spacing, since we're adding `mx-auto p-6` to the parent.

(in this line)
https://github.com/EddieHubCommunity/LinkFree/blob/ab10ce665ce4a155d07f0195f2d0d90a6adeb9d1/components/layouts/MultiLayout.js#L10

Causing something like this
image

eddiejaoude: ah good point 👍

Balastrong:
I see the main div uses `mx-auto container px-6 mt-6` and in the `Page` component we have `p-6` for the spacing... maybe we could replace that `div` with `Page` there as well, what do you think?

eddiejaoude:
It would be great *if* we can have more consistency, but I know it is not always possible, could you try and see how it looks

Balastrong:
Seems fine, the page is still displayed nicely and now we're sure all paddings are equal :)

eddiejaoude:
Great, thank you 👍

eddiejaoude:
Is it having an effect on the tabs, as I was just having a play with your branch and this looks a bit odd, do you see it also?

Screenshot 2022-12-08 at 21 15 09

Balastrong:
Oh, that's true!
I see there are some divs with `my-8`, not sure where they come from. They have always been there as I can see, but with this change they're visible 🤔

image

Balastrong:
Ah ok looks like those divs are between the tabs... do we really need them? If we remove them the page looks good.

eddiejaoude:
Oh yes, they were a temporary fix before we had the tabs, good spot, yes please remove 👍

eddiejaoude (commented):
Thank you 👍

Balastrong:
Thank you for the support! ❤

PR #2551, opened 2022-12-31
copy to clipboard feature added

### LINKED ISSUE #2521: [DOCS] Add a copy code button for all code blocks in Quickstart
### Description

To improve UX, add a copy code button/icon to the top right of each code block in the LInkfree quikstart

https://linkfree.eddiehub.io/docs/quickstart

### Screenshots

_No response_

### Additional information

_No response_

ChinmayMhatre:
## Fixes Issue
Closes #2521 

## Changes proposed

Added the copy to clipboard button to the code blocks in the quick start page.
https://linkfree.eddiehub.io/docs/quickstart

## Screenshots
image
image
image

<a href="https://gitpod.io/#https://github.com/EddieHubCommunity/LinkFree/pull/2551"></a>

### DIFF
--- a/components/ClipboardCopy.js
+++ b/components/ClipboardCopy.js
@@ -1,6 +1,7 @@
 import React, { useState } from "react";
+import { onlyText } from "react-children-utilities";
 
-const ClipboardCopy = ({ copyText, children}) => {
+const ClipboardCopy = ({children}) => {
    const [isCopied, setIsCopied] = useState(false);
 
    async function copyTextToClipboard(text) {
@@ -11,14 +12,9 @@ const ClipboardCopy = ({ copyText, children}) => {
       }
    }
 
-   // const formatText = (str) => {
-   //    const start = 3 + language.length;
-   //    return str.slice(start, -3);
-   // }
-
    const handleCopyClick = async () => {
       try {
-         await copyTextToClipboard(copyText);
+         await copyTextToClipboard(onlyText(children));
          setIsCopied(true);
          setTimeout(() => {
             setIsCopied(false);
--- a/components/ClipboardCopy.js
+++ b/components/ClipboardCopy.js
@@ -1,0 +1,51 @@
+import React, { useState } from "react";
+
+const ClipboardCopy = ({ copyText, children }) => {
+   const [isCopied, setIsCopied] = useState(false);
+
+   async function copyTextToClipboard(text) {
+      if ("clipboard" in navigator) {
+         return await navigator.clipboard.writeText(text);
+      } else {
+         return document.execCommand("copy", true, text);
+      }
+   }
+
+   const handleCopyClick = () => {
+      copyTextToClipboard(copyText)
+         .then(() => {
+            setIsCopied(true);
+            setTimeout(() => {
+               setIsCopied(false);
+            }, 1500);
+         })
+         .catch((err) => {
+            console.log(err);
+         });
+   };
+
+   return (
+      <div className="relative">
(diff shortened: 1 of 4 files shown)


components/ClipboardCopy.js
```diff
@@ -0,0 +1,51 @@
...
+   const handleCopyClick = () => {
+      copyTextToClipboard(copyText)
+         .then(() => {
+            setIsCopied(true);
+            setTimeout(() => {
+               setIsCopied(false);
+            }, 1500);
+         })
+         .catch((err) => {
+            console.log(err);
```
eddiejaoude, line 23: We are using async/await and try/catch

components/ClipboardCopy.js
```diff
@@ -0,0 +1,51 @@
+import React, { useState } from "react";
+
+const ClipboardCopy = ({ copyText, children }) => {
```
eddiejaoude, line 3: Do we need `copyText` and `children`? They look the same

eddiejaoude (changes requested):
Looks good, I left some inline questions

components/ClipboardCopy.js
ChinmayMhatre: The `children` have the code blocks commas in the start and end. Should I use slice to remove the firsts 5 and the last three characters? Then We could probably use the `children` prop itself

components/ClipboardCopy.js
ChinmayMhatre: Oh okay, I'll fix that

components/ClipboardCopy.js
eddiejaoude: Oh good point 🤔 I think yes, because I am keen to try and use the same code, so there is not duplicate and not the risk of the code being different and causing bugs for people.

It would probably be safer to separate on the line breaks and remove the first line and last line - because it might not always be 5 characters at the beginning

components/ClipboardCopy.js
eddiejaoude: thank you 👍

components/ClipboardCopy.js
ChinmayMhatre: Is there a way to get the `children` as a string?

ChinmayMhatre: Is there a way to convert the children object to string?

ChinmayMhatre: > Oh good point 🤔 I think yes, because I am keen to try and use the same code, so there is not duplicate and not the risk of the code being different and causing bugs for people.
> 
> It would probably be safer to separate on the line breaks and remove the first line and last line - because it might not always be 5 characters at the beginning

Oh yeah, That is a way better option. Could you tell me how would I convert the `children` prop to string.  I can't seem to find the method to get the text from the object

eddiejaoude: I haven't tried before, but after a Google I found this https://github.com/fernandopasik/react-children-utilities

ChinmayMhatre: I'll give it a go, thanks!

ChinmayMhatre:
Hey Eddie! I fixed it 🕺

eddiejaoude:
Looks good @ChinmayMhatre 👍 I will have a play with it later on and then we can get it merged

Thank you!

ChinmayMhatre:
> Looks good @ChinmayMhatre 👍 I will have a play with it later on and then we can get it merged
> 
> Thank you!

Awesome! Maybe later on we could move the wrapper component to a utils folder. Could create an issue later on, a green square for a first timer.

eddiejaoude (approved):
Looks great 👍

PR #2577, opened 2023-01-02
Imtiazpy patch 1

imtiazpy:


eddiejaoude (changes requested):
It looks like you have conflicts with yourself, this could be because you didn't update your fork and branch before making more changes

Your file already exists https://github.com/EddieHubCommunity/LinkFree/blob/main/data/imtiazpy.json

Issue #2581, opened 2023-01-02
Popular profiles should not have profiles that do not display views


eddiejaoude:
### Description

If a profile has stats hidden, then we should not include their profile on the popular profile page

The API `api/users/popular` will need to only return profiles that have stats enabled for their profile

KlausMikhaelson:
Hey @eddiejaoude can I work on this issue ?

eddiejaoude:
Sure @KlausMikhaelson

eddiejaoude:
@KlausMikhaelson the issue is still for you, @shyam0705 didn't notice it was assigned and kindly closed their pull request

KlausMikhaelson:
thanks you @eddiejaoude will create the pr soon !

KlausMikhaelson:
Hey @eddiejaoude have created the pr, do lmk if there is anything to change !
Thank you !

KlausMikhaelson:
Hey @eddiejaoude was facing some issues with my forked repo so I reforked it and now instead of filtering the first 10 profiles it's checking the stats first and than out of those popular profiles who has there stats public it's gonna take the first 10.
took the reference from shaym so thank you !

PR #2655, opened 2023-01-04
removed profiles from popular whose stats are not public

KlausMikhaelson:
## Fixes Issue
#2581 

## Changes proposed
it will remove those profiles whose stats are not public

- [x ] My code follows the code style of this project.
- [x ] This PR does not contain plagiarized content.
- [ x] The title of my pull request is a short description of the requested changes.

## Screenshots
image

## Note to reviewers
Have made the changes now instead of filtering the profiles initially it will do that after checking if the profiles have status public or not and after that if they are public it will show the profiles and from those profiles it will slice the first 10
Do lmk if there is anything to change !
Thank you !

<a href="https://gitpod.io/#https://github.com/EddieHubCommunity/LinkFree/pull/2655"></a>

### DIFF
--- a/pages/api/users/popular.js
+++ b/pages/api/users/popular.js
@@ -7,7 +7,7 @@ import Profile from "../../../models/Profile";
 export default async function handler(req, res) {
   await connectMongo();
 
-  const getProfiles = await Profile.find({}).sort({ views: -1 }).limit(10);
+  const getProfiles = await Profile.find({}).sort({ views: -1 });
 
   // check for db results
   if (getProfiles.length === 0) {
@@ -29,12 +29,13 @@ export default async function handler(req, res) {
         };
       }
 
-      return { ...user, username: profile.username };
+      return [];
     } catch (e) {
       console.log(`ERROR loading profile "${filePath}"`);
       return [];
     }
   });
 
-  res.status(200).json(profiles);
+  const sortedProfiles = profiles.slice(0, 10)
+  res.status(200).json(sortedProfiles);
 }
--- a/pages/api/users/popular.js
+++ b/pages/api/users/popular.js
@@ -7,7 +7,7 @@ import Profile from "../../../models/Profile";
 export default async function handler(req, res) {
   await connectMongo();
 
-  const getProfiles = await Profile.find({}).sort({ views: -1 });
+  const getProfiles = await Profile.find({}).sort({ views: -1 }).limit(50);
 
   // check for db results
   if (getProfiles.length === 0) {


eddiejaoude:
Btw you don't need to close an open PR to add more changes, you can push new commits to the same branch in your fork and GitHub will update the open PR

pages/api/users/popular.js
```diff
@@ -7,7 +7,7 @@ import Profile from "../../../models/Profile";
 export default async function handler(req, res) {
   await connectMongo();
 
-  const getProfiles = await Profile.find({}).sort({ views: -1 }).limit(10);
+  const getProfiles = await Profile.find({}).sort({ views: -1 });
```
eddiejaoude, line 10: I think this will bring back too many results, currently about 600 and in the future even more.

KlausMikhaelson: > Btw you don't need to close an open PR to add more changes, you can push new commits to the same branch in your fork and GitHub will update the open PR

Ooo okay, didn't know about it earlier, will make sure to do that next time !

eddiejaoude (changes requested):
I think binging back all the data is too much, it will load many files.

This is a challenging problem to solve. I think it might need paginated recursion to achieve this, but I am open to other ideas

pages/api/users/popular.js
KlausMikhaelson: > I think this will bring back too many results, currently about 600 and in the future even more.

so maybe making it like .limit(100) and than it will check from the first 100 and sort them out as per the stats are public or not will work ?

KlausMikhaelson: but still if more than 90 people in the top 100 has their stats hidden than it won't be good ig

eddiejaoude: We dot know if it will work or not because it depends which are the popular profiles and if their stats are public - this can always change also there is no way of knowing

I shared my thoughts in the main part of the PR

> This is a challenging problem to solve. I think it might need paginated recursion to achieve this, but I am open to other ideas

KlausMikhaelson:
> I think binging back all the data is too much, it will load many files.
> 
> This is a challenging problem to solve. I think it might need paginated recursion to achieve this, but I am open to other ideas

I don't know what it is but will check it out right away !

pages/api/users/popular.js
shyam0705: > I think binging back all the data is too much, it will load many files.
> 
> This is a challenging problem to solve. I think it might need paginated recursion to achieve this, but I am open to other ideas

But in the worst case, we have to fetch all profiles in that approach as well.

eddiejaoude:
Sure, I think it would need to collect `20` in a function with pagination, then each time it has less than 10 at the end, it calls the function again to get the next 20 paginate (recursively) - it is a bit of a tricky situation

Maybe for now we keep it simple and select a large number like 50 for now and another PR later can come and improve it further - the worse case if less than 10 come back, I think the UI will just show less results

KlausMikhaelson:
Ooo okay got the idea, so should I make the changes now i.e. taking the 50 first?,
got the idea of recursive pagination but haven't done it so will check it out and hopefully will create a pr for it too in future !

eddiejaoude:
yes I think the DB call should take 50 records and then slice at the end to always return a max of 10

KlausMikhaelson:
awesome ! have made the changes !

eddiejaoude (approved):
Looks good 👍

Issue #2243, opened 2022-11-18
Release action uses unmaintained actions/create-release


dan-mba:
### Description

The GitHub action [actions/create-release](https://github.com/actions/create-release) is no longer maintained.

The repo lists 4 maintained actions to consider switching to.

Rishav-12:
Hey, can I try working on this?

krshkun:
I will implement `semantic-release`

SaraJaoude:
@krshkun I am going to unassign from you this as there has been no activity

krshkun:
`semantic-release`'s new version does not support v16 node (the one we use), it must be above v18, we had a conversation about migrating all actions to lts (v18). Awaiting that one to be able to get `semantic-release` in otherwise it will break all current actions.

Sorry for not updating it here.

eddiejaoude:
> semantic-release's newer versions do not support v16 node (the one we use), it must be above v18, we had a conversation about migrating all actions to lts (v18)

I think other Actions can be updated later, only this Action needs node v18

krshkun:
But all the other actions will fail on npm install because of the difference? :thinking:

Pradumnasaraf:
I am working on this issue.

PR #2808, opened 2023-01-07
Change workflows to not push to main directly.

### LINKED ISSUE #2756: Main branch is not protected because of the GH Action needs to write
### Description

The default branch `main` should be protected but for the Actions that need to write to `main` this is a problem

What are the solutions?

### Screenshots

_No response_

### Additional information

_No response_

mocdaniel:
## Fixes Issue

Closes #2756 

## Changes proposed

[X] Changes workflow `releases.yml` to not push to `main` directly. More info below.
[] Changes workflow `prettier.yml` to not push to `main` directly. More info below.

## Note to reviewers

The proposed changes to the actions which are currently pushing to `main` follow the same pattern:

1. Instead of pushing to main, push to a uniquely named, new branch, e.g. `release-from-<commit-sha>`.
2. Create a PR using the `gh` CLI tool which is installed and usable on all managed GitHub runners (see [here](https://docs.github.com/en/actions/using-workflows/using-github-cli-in-workflows)).
3. Merge this PR using the `gh` CLI tool and a token which authorizes the action to do the merge. In the current draft, a merge commit is used and the merge gets forced, i.e. pending/failed actions are ignored via the `--admin` flag.
4. Continue with the remaining, existing steps, as all changes are now once again in `main`.
5. Be happy about green squares and a neat collaboration experience. 🥳 

@eddiejaoude @loftwah what do you think about this proposed strategy?

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -17,6 +17,20 @@ jobs:
         with:
           github-token: ${{ secrets.CHANGELOG_RELEASE }}
           version-file: "./package.json,./package-lock.json"
+          git-branch: "release-from-${{ github.sha }}"
+
+      - name: create PR with release info
+        id: create-pr
+        run: gh pr create -B main -H release-from-${{ github.sha }} --title 'Merge new release into main' --body 'Created by Github action'
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: merge PR with release info
+        id: merge-pr
+        run: git checkout release-from-${{ github.sha }} && gh pr merge --admin -A <eddiehub mail> --merge --subject 'Merge release info' --delete-branch
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+
 
       - name: create release
         uses: actions/create-release@v1


.github/workflows/release.yml
```diff
@@ -17,6 +17,20 @@ jobs:
         with:
           github-token: ${{ secrets.CHANGELOG_RELEASE }}
           version-file: "./package.json,./package-lock.json"
+          git-branch: "release-from-${{ github.sha }}"
+
+      - name: create PR with release info
+        id: create-pr
+        run: gh pr create -B main -H release-from-${{ github.sha }} --title 'Merge new release into main' --body 'Created by Github action'
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
loftwah, line 26: Do we need to state this explicitly? GitHub Actions should have access to GitHub CLI without us explicitly giving it a token, but I've had to do it this way before even with that.

mocdaniel: I don't know. I think at least for the part where we merge the new branch, we'll need a token with elevated permissions. I will spin up a test-repository and try the action, will link the results here.

dan-mba:
I think since the release action will be modifying the changelog file on main it might have issues as well.
The action currently used to create the release is unmaintained, so it could present a problem.
See issue #2243

.github/workflows/release.yml
mocdaniel: I did some testing and all the actions do in fact only work when the token is provided explicitly.

mocdaniel:
I worked it out, I think. I commented all involved steps for better documentation of what is actually happening, see also this action log as proof of concept:

[https://github.com/mocdaniel/action-test/actions/runs/3862144796/jobs/6583557437](https://github.com/mocdaniel/action-test/actions/runs/3862144796/jobs/6583557437)

.github/workflows/release.yml
```diff
@@ -9,20 +9,57 @@ jobs:
...
+        run: git fetch --all && git checkout --track origin/release-from-${{ github.sha }}
+
+
+      # create release info and push it upstream
       - name: conventional Changelog Action
         id: changelog
         uses: TriPSs/conventional-changelog-action@v3
         with:
-          github-token: ${{ secrets.CHANGELOG_RELEASE }}
+          github-token: ${{ secrets.GITHUB_TOKEN }}
```
eddiejaoude, line 37: I think we will need to switch this back before merging, otherwise other Actions won't run

eddiejaoude:
Btw your [test 2](https://github.com/mocdaniel/action-test/pull/6) PR has conflicts, is that for another reason?

.github/workflows/release.yml
loftwah: Bah! Same problem that happened to me recently. The docs are a lie 😂😅

Panquesito7: > Bah! Same problem that happened to me recently. The docs are a lie 😂😅

Kinda off-topic: you can submit an issue/PR in the [`docs`](https://github.com/github/docs) repository about it. 🙂

mocdaniel:
> Btw your [test 2](https://github.com/mocdaniel/action-test/pull/6) PR has conflicts, is that for another reason?

Yes, that's because several 'release branches' exist(ed) at the same time due to the action not working at that point of time. Now that everything works, there will never be competing 'release branches', so this shouldn't be a problem, really.

.github/workflows/release.yml
mocdaniel: Done.

mocdaniel: After reading up on the matter once more, I am sure we will need to use `GITHUB_TOKEN` for every step except the one creating the actual release within GitHub. That step will use `CHANGELOG_RELEASE` in order to trigger `deploy.yml`.

This is necessary in order to prevent endless CI circles from happening, as we create and merge PRs to main in the process, which would trigger another instance of the same job if we didn't use `GITHUB_TOKEN`.

See also [https://docs.github.com/en/actions/using-workflows/triggering-a-workflow#triggering-a-workflow-from-a-workflow](https://docs.github.com/en/actions/using-workflows/triggering-a-workflow#triggering-a-workflow-from-a-workflow).

mocdaniel:
This combination of workflows does its job now, but for some reason I couldn't figure out it created empty changelogs for me. We might have to test this a bit more, or maybe one of you knows why this is happening.

.github/workflows/prettier.yml
```diff
@@ -1,25 +1,48 @@
 name: Format
 on:
   push:
-    paths:
-      - "data/**"
+    branches:
+      - main
```
dan-mba, line 5: Would it make sense to have both of these?

Prettier only changes the files in the data directory,
so only run it on changes to the data directory pushed to main.

mocdaniel: Unfortunately, adding several triggers to a workflow chains them as an `OR` expression, not an `AND` expression. Therefore, we can't use this approach, because then we would trigger the `prettier.yml` workflow for pushes to arbitrary branches.

dan-mba: I use the following on one of my repos:
```yml
  push:
    branches:
      - main
    paths-ignore:
      - ".github/**"
 ```

Is the reason this works because it is `paths-ignore`?

mocdaniel: You're right, this does work. `branches` and `paths`/`paths-ignore` in combination actually get interpreted as an `AND` expression...

I'd rather leave it like this though, because now the `release.yml` workflow depends on `prettier.yml` so we want `prettier.yml` to run for every push to `main` in order to check whether we can cut a new release afterwards.

dan-mba: OK

PR #2848, opened 2023-01-08
data: alchemicalartist

alchemicalartist:


data/AlchemicalArtist.json
```diff
@@ -0,0 +1,24 @@
...
+  "name": "Chris Denton",
+  "displayStatsPublic": false,
+  "type": "personal",
+  "bio": "artist, musician, human",
+  "avatar": "https://GitHub.com/alchemicalArtist.png",
+  "links": [
+    {
+      "name": "Subscribe on YouTube!",
+      "url": "https://YouTube.com/@AlchemicalArtist",
+      "icon": "faYoutube"
```
eddiejaoude, line 11: The icons should have a capital `Fa`, so it would be `FaYoutube`, you can check the icon names here in the docs https://linkfree.eddiehub.io/icons

data/AlchemicalArtist.json
```diff
@@ -0,0 +1,24 @@
+{
+  "name": "Chris Denton",
+  "displayStatsPublic": false,
+  "type": "personal",
+  "bio": "artist, musician, human",
+  "avatar": "https://GitHub.com/alchemicalArtist.png",
```
eddiejaoude, line 6: Your username is not the exact same case as your GitHub username, this will work now but might not in the future, it is best to keep it exactly the same

Please do the same with the file name also, have it match your GitHub username exactly

eddiejaoude (changes requested):
Looks good, just a few small tweaks

eddiejaoude (approved):
Looks good 👍

PR #3169, opened 2023-01-11
User map

amandamartin-dev:
This is a POC to test and idea to be able to see the locations of linkfree users.  So far added the following:
react-leaflet
leaflet
simple map component
map page

- [x] where to put nav button - this will replace the `Discover nav button https://github.com/EddieHubCommunity/LinkFree/pull/4642

## Screenshots

Screenshot 2023-02-09 at 10 55 45

## Note to reviewers

<a href="https://gitpod.io/#https://github.com/EddieHubCommunity/LinkFree/pull/3169"></a>


eddiejaoude:
Looks great @amandamartin-dev 👍 

I added a function to get the user's GitHub location and add their geo location to the profile in the db

amandamartin-dev:
Oh nice! I was planning on getting back to this on the weekend. Will def check that out!

amandamartin-dev:
@eddiejaoude I noticed something potentially unusual.  I see you added the users and <DynamicMap /> to the search page. Was this intentional?  Want to make sure I'm working in the correct location/understand the concept.

amandamartin-dev:
I could see adding this to the search page but perhaps giving the option to see profile cards or the map view as both at the saem time would be cluttered.  Would also have to handle the map not rendering until after the search is complete perhaps as the "flashing" on the fuzzy/dynamic search is not ideal.  Lmk your thoughts on where the map would "live"

eddiejaoude:
Oh great points Amanda! Yes let's keep it simple for now 👍 let's remove it from the search page

amandamartin-dev:
@eddiejaoude playing around a but more, I noticed the api you added is pulling 1 location at a time by username.  If I want to adjust this to include location in all of the profiles (similar to where hydrateWithStats) is adding all the stats, are there any issues with modifying the code to update all the profiles?  I'm worried about hitting the GitHub API that much. 

What I'm thinking for the POC would be:
On map load all eddiehub profiles are plotted (marker cluster will be used)
when a user clicks on a map point, the popup would show some profile data

Later improvement - integration with filter/search ability

So...i think i need to load location data (where available) to all profile JSON.  Wdyt?

amandamartin-dev:
@eddiejaoude another random thoguht about this just came to my head...what if instead of trying to add ALL locations every time we run the code, automatically adding user location to the JSON on new profile addition so it's always available in the data.  Then we could backfill all the current ones and not have to worry about API calls as linkfree scales. 

caveat - how to handle location updates could be tricky this way

eddiejaoude:
I was still thinking how best to get/update everyone's location because we don't want to hit the GitHub API rate limit.

Yep keep let's simple to begin with and that we can launch it sooner, we can add search/filter etc later 👍

eddiejaoude:
I think we can improve the location "update" later, your idea of doing it when the profile is added is a great idea 👍

amandamartin-dev:
> I was still thinking how best to get/update everyone's location because we don't want to hit the GitHub API rate limit.

I feel like maybe @dan-mba might have an idea here. Thoughts on how best to get all the linkfree profiles loaded with locations without overloading the GH API?

If not, I think I'll drop this Q in discord bc I'm not sure right now but someone surely will have another idea

amandamartin-dev:
> I think we can improve the location "update" later, your idea of doing it when the profile is added is a great idea 👍

Okay so then to backfill, would just need to create and batch run a script on existing profiles to not hit the rate limits maybe??  Want to tread carefully lol

eddiejaoude:
> User-to-server requests are limited to 5,000 requests per hour 

I think we will be fine if we do it authenticated API request, I can add a PAT on the server as an environment variable - we have just over 1k users so should be within that limit

amandamartin-dev:
> I think we will be fine if we do it authenticated API request, I can add a PAT on the server as an environment variable - we have just over 1k users so should be within that limit

Amazing! I will need to be pointed in the correct direction to know where to run a one time update script for the database (unless you prefer to do that)

And then for the new profiles added, I'm assuming a github action can query and add location on merge? Haven't checked that yet so it's still an assumption

eddiejaoude:
Sure, I am happy to work on the API part 👍

amandamartin-dev:
Sweet! Tag me once there's some data ready to go and I can finish up the map. Figuring out the best route for the data backfill was really the trickiest part 😂

No rush at all obvi

eddiejaoude:
I think for now during development, you can get some data but visiting some profiles locally and it will fetch those profiles location from GitHub's API - if you do 10 profiles you shouldn't hit the rate limit - that way it isn't a blocker for you? (or am I missing something?)

amandamartin-dev:
Oh true! I can modify the code a bit to just save it in the object. So far it wasn't ever returning location in the profiles but I didn't debug that yet since I was thinking I needed location for all vs one at a time.

tl;Dr yes, I can keep moving on it

dan-mba:
Why are all these testimonials being committed into this branch?

dan-mba:
@eddiejaoude Any Idea why I can't use GitHub's GraphQL Explorer with EddieHubCommunity data?

When I run this query
```graphql
{
  organization(login: "EddieHubCommunity") {
    memberStatuses(first: 10) {
      nodes {
        user {
          location
          login
        }
      }
    }
  }
}
```

I get this response
```json
{
  "data": {
    "organization": null
  },
  "errors": [
    {
      "type": "FORBIDDEN",
      "path": [
        "organization",
        "memberStatuses"
      ],
      "extensions": {
        "saml_failure": false
      },
      "locations": [
        {
          "line": 3,
          "column": 5
        }
      ],
      "message": "Although you appear to have the correct authorization credentials, the `EddieHubCommunity` organization has enabled OAuth App access restrictions, meaning that data access to third-parties is limited. For more information on these restrictions, including how to enable this app, visit https://docs.github.com/articles/restricting-access-to-your-organization-s-data/"
    }
  ]
}
```

eddiejaoude:
> Why are all these testimonials being committed into this branch?

not sure what happened when I updated the branch, I think I have fixed it now

eddiejaoude:
I haven't tried the `GitHub's GraphQL Explorer` so I am not sure. @Cahllagerfeld I think might have? I use the REST api

dan-mba:
> I haven't tried the `GitHub's GraphQL Explorer` so I am not sure. @Cahllagerfeld I think might have? I use the REST api

The question I had was related to the error message

the `EddieHubCommunity` organization has enabled OAuth App access restrictions, meaning that data access to third-parties is limited.

amandamartin-dev:
@eddiejaoude (or anyone interesting in continuing)

I've hardcoded my username in the API call to get location and built out the map marker based on that.
There are some todos in the map code to consider once we have more data coming in

Here's the poc with my user
POC of a leaflet map in linkfree with one profile marker loaded and popup expanded

eddiejaoude:
> The question I had was related to the error message

> the EddieHubCommunity organization has enabled OAuth App access restrictions, meaning that data access to third-parties is limited.

Sorry @dan-mba I will take a look at the Org's settings

eddiejaoude:
Looks great @amandamartin-dev 🔥

amandamartin-dev:
@eddiejaoude i just merged in main and fixed some merge conflicts.  When i pushed there's a random profile in data/ included with my merge that's causing a failure.  No CLUE what happened, but I know this happened to you. What fixed it?

eddiejaoude:
It happens sometimes, I would recommend trying to revert the file back to `main` with `git checkout main -- <filename>`

dan-mba:
> @eddiejaoude i just merged in main and fixed some merge conflicts.  When i pushed there's a random profile in data/ included with my merge that's causing a failure.  No CLUE what happened, but I know this happened to you. What fixed it?

I think this is caused by the prettier git hook.

To fix it, run the workflow dispatch on the format action, then merge main into your branch.

amandamartin-dev:
> run the workflow dispatch on the format action

I'm not familiar with using github actions, so I'm not sure where I would run that.  From the command line or in here? 

Also, should I wait? Meaning - if it's going to happen every time I can just ignore the errors for now as they aren't harming me til if/when we actually want to use this branch. wdyt?

dan-mba:
> > run the workflow dispatch on the format action
> 
> 
> 
> I'm not familiar with using github actions, so I'm not sure where I would run that.  From the command line or in here? 
> 
> 
> 
> Also, should I wait? Meaning - if it's going to happen every time I can just ignore the errors for now as they aren't harming me til if/when we actually want to use this branch. wdyt?
> 
> 
> 
> EDIT - i just saw eddies suggestion too so i can try that as well

To run the format action, click the run workflow button at this link

https://github.com/EddieHubCommunity/LinkFree/actions/workflows/prettier.yml

amandamartin-dev:
@dan-mba omg! you def taught me something new.  I don't mess with the actions since I haven't written any before and so far when they have errored, it's been code related/obvious to fix so i didn't know about re-running them safely.  Thanks!

eddiejaoude:
Something we should probably document 👍

dan-mba:
> Something we should probably document 👍 

If there is a maintainer only documentation, I would agree.
I would rather avoid suggesting to run actions to new contributors.

eddiejaoude:
> If there is a maintainer only documentation, I would agree.

@SaraJaoude has started a maintainer docs (faqs) - we should put this out asap and then it can get feedback/improved

amandamartin-dev:
current status: ability to test with api data, standard layout items added

note: there's no unique keys for the data in the map, map marker is commented, uncomment to work

needs: perm api solution, solution for no keys in the json files(need a unique key for the map), add marker cluster

amandamartin-dev:
@eddiejaoude 
The POC with testing data is ready for hooking up the actual API but I wasn't sure how to proceed.  We discussed further up in the thread backfilling the exiting db with location values (where available) and then adding a trigger to new profiles in the future to automatically query location and add to the db .

Should we open issues against the live repo to do that first then I can rework the map code to remove the API calls after it's live?

Here's a video of the test data and feel free to play around
You will notice in the video that not ALL data is displayed.  this is due to not all users having a location and I put a check in the map code in case any data gets through without location so it wont break the map

https://user-images.githubusercontent.com/97615019/215926826-f67a6f47-2d3b-47d6-bd0a-1d821de3d597.mov

eddiejaoude:
I think this is ready to be reviewed 🎉 

Note: there is no link to maps just yet, but we can still start querying people's location from GitHub if they have it public

eddiejaoude:
Thank you @amandamartin-dev 👍 - I will add the nav button after this PR gets merged https://github.com/EddieHubCommunity/LinkFree/pull/4642

eddiejaoude:
btw I have resolved conflicts, just doing some testing and I will push shortly

amandamartin-dev:
Okay! You have some time. I'll likely be back at my computer by 9ET

amandamartin-dev:
Documenting here for continuity what i just put in discord.

Map looks great, I pushed one commit to disable zoom on scroll as it was annoying UX with the map full page if you can use that update for the final merge.

Question - the map does not appear to populate data until each profile has been clicked on once to trigger the query.  Is there a way to trigger them without this?  If not, the map will definitely be confusing for awhile

Subsequent bug ticket to open:
Handle situation where location is populated, but not "real" which will default place a point in the ocean. Example profile that triggered this was for Barsha Das whose github location is global/remote.  Can put in a check for this later as it is not critical to functionality.

Final note - avatars were populating very slowly, but I attribute this to the beta link that was deployed for testing. 

Let me know if I should check anything else at this point!

eddiejaoude:
Thanks @amandamartin-dev 

> Question - the map does not appear to populate data until each profile has been clicked on once to trigger the query. Is there a way to trigger them without this? If not, the map will definitely be confusing for awhile

I was thinking most profiles will get a visit, and I couldn't find a better way to get the location without getting rate limited

> Subsequent bug ticket to open:
Handle situation where location is populated, but not "real" which will default place a point in the ocean. Example profile that triggered this was for Barsha Das whose github location is global/remote. Can put in a check for this later as it is not critical to functionality.

I have just added a fix for that, `unknown` locations will be ignored and not returned by the API

> Final note - avatars were populating very slowly, but I attribute this to the beta link that was deployed for testing.

oh, that is something we will have to keep an eye on. Beta server is smaller but not sure how that would effect the images

amandamartin-dev:
That makes sense about the query. I think we'll see once it's live how things go and iterate from there. Only other option I can think of is storing location in the db instead of cache and updating on profile visits, but can be a future idea if needed

eddiejaoude:
> Only other option I can think of is storing location in the db instead of cache and updating on profile visits, but can be a future idea if needed

Sorry by "cached", I meant it is stored in the db but we revalidate it after 7 days (this can be changed easily to make longer/shorter if required

amandamartin-dev:
> Sorry by "cached", I meant it is stored in the db but we revalidate it after 7 days (this can be changed easily to make longer/shorter if required

omg, haha! I thought cache was an unusual choice.  I think that's fine as the dataset will grow over time.  I thoght you were grabbing a data point and it expired after 7 days LOL.  I think this will work fine

PR #3246, opened 2023-01-12
feat: api handle redirects

### LINKED ISSUE #3242: [FEATURE] First step into API
### Description

https://linkfree.eddiehub.io/api/links/handle?url=https://github.com/eddiejaoude&user=eddiejaoude

ofc the http thing should be uri encoded via encodeUriComponent function

And then the link even is recorded when this happens. This will be a good start considering we are building a API 👍 

### Screenshots

_No response_

### Additional information

_No response_

krshkun:
## Fixes Issue

closes #3242 

## Note to reviewers

<a href="https://gitpod.io/#https://github.com/EddieHubCommunity/LinkFree/pull/3246"></a>


components/user/UserLink.js
```diff
@@ -11,28 +11,12 @@ export default function UserLink({
...
-    } catch (e) {
-      // TODO: link not found
-      console.log("ERROR link not found ", e);
-    }
-  };
 
   return (
     <a
-      href={link.url}
+      href={`${BASE_URL}/api/handler/links?username=${username}&url=${encodeURIComponent(link.url)}`}
```
loftwah, line 17: Should there be quotes here?

krshkun: Hmm, which quotes? there are already backticks

components/user/UserSocials.js
```diff
@@ -2,31 +2,17 @@ import getIcon from "../Icon";
...
-      // TODO: link not found
-      console.log("ERROR link not found ", e);
-    }
-  };
 
   return (
     <a
-      href={social.url}
-      onClick={clickLink}
+      href={`${BASE_URL}/api/handler/links?username=${username}&url=${encodeURIComponent(
```
loftwah, line 8: They're also present here.

krshkun: I don't know why there should be quotes. I've ran some manual tests and the link works 👍

tests/profile.spec.js
```diff
@@ -27,24 +27,11 @@ test("Profile views increase", async ({ page }) => {
...
-
-  await page.waitForResponse(
-    (resp) => resp.url().includes("/api/statistics") && resp.status() === 201
-  );
-
-  const endingClicks = (await link.innerText()).match(/(\d+)/g);
-
-  // 4. get the current link views and see if increased by 3
-  expect(parseInt(startingClicks[1])).toEqual(parseInt(endingClicks[1]) - 3);
+test.fixme("Link clicks increase", async ({ page }) => {
```
krshkun, line 30: I am commenting out tests, why? Because this API completely changes that test and it has to be completely rewritten

krshkun: I will make the issue once the PR has been merged 👍

eddiejaoude: We have incomplete tests that need writing but we don't want to remove existing tests otherwise we end up going backwards not forwards - please include existing tests in the PR also

krshkun: Tests added

tests/profile.spec.js
```diff
@@ -27,24 +27,24 @@ test("Profile views increase", async ({ page }) => {
   expect(parseInt(startingViews)).toEqual(parseInt(endingViews) - 3);
 });
 
-test("Link clicks increase", async ({ page }) => {
+test.fixme("Link clicks increase", async ({ page }) => {
```
eddiejaoude, line 30: I think with `.fixme` it won't run?

eddiejaoude (commented):
Looks good 👍 , I will have a manual test first thing tomorrow morning

tests/profile.spec.js
krshkun: Now I know why it was passing 😂

krshkun: Fixed it, tests still pass 👍

pages/api/handler/links.js
```diff
@@ -0,0 +1,119 @@
...
+        clicks: 1,
+      });
+    } catch (e) {
+      const err = `ERROR incrementing platform stats: ${e}`;
+      console.log(err);
+      return res.res(500).json({ message: err });
+    }
+  }
+
+  return res.status(201).redirect(decodeURIComponent(url));
```
eddiejaoude, line 118: This file looks very similar to the existing api handler file `pages/api/statistics/[...data].js`, we don't need both especially now as the other one doesn't seem to be used anymore.

I prefer the other one because it handles the uri as part of the path rather than optional query parameters.

Could we add this redirect line to the existing file `pages/api/statistics/[...data].js`? https://github.com/EddieHubCommunity/LinkFree/blob/main/pages/api/statistics/%5B...data%5D.js

Or am I missing something?

krshkun: Yes, but that will be naming things badly, because now it does not just handle the stats, but also the link, the process of the link. Also the query parameters aren't optional, you give the query parameters you get an answer, if you don't, you get a error. I prefer this approach because if a person goes to just api/statistics they get a 404 code, or they go to just api/statistics/:username, the still just get a 404 code. Parameters will come useful incontext to the coming API 
image

eddiejaoude: I know you have checked for those parameters, but from what I remember for REST, query parameters are usually for optional ones (regardless if you check for them or not), where when it is part of the path, it is for the resource

For example a filter would be a query parameter (like keyword or sort) but the resource of what is being filter (like users or posts) would be part of the path.

eddiejaoude: Maybe it shouldn't be on `statistics`, but I still think it should be part of the path, similar to how it is now

I think the best option is to add the redirect to the existing file, we don't need unused code, and move that file out of `statistics`

eddiejaoude:
Btw one downside I noticed by doing this server side is that the numbers are not incremented on the user's page - but I don't think this is a blocker

pages/api/handler/links.js
krshkun: So you mean deleting the file statistics and making it a part of the path? 👍

krshkun:
@eddiejaoude We can increment that but that will cause the component to rerender like it doing previously. if you are fine with that we can add a state and onClick function

pages/api/handler/links.js
eddiejaoude: Probably better to use that file and just move it instead of deleting it

eddiejaoude:
Hmm maybe, as it is a better user experience

krshkun:
Client-side only. Doesn't make sense to fetch again since we will already have the click count from when we are displaying it for the first time

pages/api/users/[username]/links/[url].js
```diff
@@ -0,0 +1,121 @@
...
+      });
+    } catch (e) {
+      const err = `ERROR incrementing platform stats: ${e}`;
+      console.log(err);
+      return res.res(500).json({ message: err });
+    }
+  }
+
+  return res.status(201).redirect(decodeURIComponent(url));
+}
```
Comment (eddiejaoude) [inline pages/api/users/[username]/links/[url].js:121]: There are only a few lines difference with the file in statistics, less changes are more 😉 

It would be better to add the few tweaks to the original file and move it, rather than creating a new duplicate file 
https://github.com/EddieHubCommunity/LinkFree/blob/main/pages/api/statistics/%5B...data%5D.js

Comment (krshkun) [reply pages/api/users/[username]/links/[url].js:121]: No no no. I've done exactly that I just forgot to push deleting the statistics file, did that now

Comment (krshkun) [reply pages/api/users/[username]/links/[url].js:121]: I've improved that file itself

Comment (eddiejaoude) [reply pages/api/users/[username]/links/[url].js:121]: @krshkun I did a diff there are only a few line changes please keep it simple...

```diff
diff --git a/pages/api/statistics/[...data].js b/pages/api/statistics/[...data].js
index 7e05e900..47c0d572 100644
--- a/pages/api/statistics/[...data].js
+++ b/pages/api/statistics/[...data].js

 export default async function handler(req, res) {
   await connectMongo();
-  const { data } = req.query;
-  const username = data[0];
-  const url = data[1];
 
-  if (req.method != "PUT" || !username || !url) {
+  const { username, url } = req.query;
+
+  if (req.method != "GET" || !username || !url) {
     return res
       .status(400)
-      .json({ error: "Invalid request: 'username' and url 'required'" });
+      .json({ error: "Invalid request: username and link required" });
   }
 
-  const getProfile = await Profile.findOne({ username });
+  let getProfile = await Profile.findOne({ username });
 
   if (!getProfile) {
-    return res
-      .status(404)
-      .json({ error: "Invalid request: 'username' not found" });
+    try {
+      await Profile.create({
+        username,
+        views: 1,
+      });
+
+      getProfile = await Profile.findOne({ username });
+    } catch (e) {
+      const err = `ERROR creating profile: ${e}`;
+      console.log(err);
+      return res.status(500).json({ message: err });
+    }
   }
 
   const getLink = await Link.findOne({ username, url });
+
   if (getLink) {
     try {
       await Link.updateOne(
@@ -36,7 +48,9 @@ export default async function handler(req, res) {
         }
       );
     } catch (e) {
-      console.log("ERROR incrementing link", e);
+      const err = `ERROR incrementing link stats: ${e}`;
+      console.log(err);
+      return res.res(500).json({ message: err });
     }
   }
 
@@ -48,6 +62,7 @@ export default async function handler(req, res) {
         url,
         clicks: 1,
       });
+
       await Profile.updateOne(
         {
           username,
@@ -58,13 +73,18 @@ export default async function handler(req, res) {
         { new: true, useFindAndModify: false }
       );
     } catch (e) {
-      console.log("ERROR creating link", e);
+      const err = `ERROR creating link: ${e}`;
+      console.log(err);
+      return res.res(500).json({ message: err });
     }
   }
 
   const date = new Date();
+
   date.setHours(1, 0, 0, 0);
+
   const getPlatformStats = await Stats.findOne({ date });
+
   if (getPlatformStats) {
     try {
       await Stats.updateOne(
@@ -76,7 +96,9 @@ export default async function handler(req, res) {
         }
       );
     } catch (e) {
-      console.log("ERROR incrementing platform stats", e);
+      const err = `ERROR incrementing platform stats: ${e}`;
+      console.log(err);
+      return res.res(500).json({ message: err });
     }
   }
 
@@ -89,10 +111,11 @@ export default async function handler(req, res) {
         users: 0,
       });
     } catch (e) {
-      console.log("ERROR creating platform stats", e);
+      const err = `ERROR incrementing platform stats: ${e}`;
+      console.log(err);
+      return res.res(500).json({ message: err });
     }
   }
 
-  const latestLink = await Link.findOne({ username, url });
-  res.status(201).json(latestLink);
+  return res.status(201).redirect(decodeURIComponent(url));
 }
```

Then we can see the exact changes and discuss the specific line changes more easily

Comment (krshkun) [reply pages/api/users/[username]/links/[url].js:121]: I'm not really sure what do you want me to do in the PR regarding the API requests, every time, even if I revert everything, just have the [...data].js file on the upstream and move it to my directory, it shows them as two separate changes. If you want to try this out, just open `EddieHubCommunity/LinkFree:main` and change the file's name and commit or something

Comment (krshkun) [reply pages/api/users/[username]/links/[url].js:121]: This is the diff

```diff
1,5c1,8
--- pages/api/statistics/[...data].js	2023-01-14 21:18:09.523506221 +0530
+++ pages/api/users/[username]/links/[url].js	2023-01-14 21:14:51.613508229 +0530
@@ -1,29 +1,41 @@
-import Link from "../../../models/Link";
-import Profile from "../../../models/Profile";
-import Stats from "../../../models/Stats";
-import connectMongo from "../../../config/mongo";
-
+import Link from "../../../../../models/Link";
+import Profile from "../../../../../models/Profile";
+import Stats from "../../../../../models/Stats";
+import connectMongo from "../../../../../config/mongo";
+
+/**
+ * @type {import('next').NextApiHandler}
+ */
 export default async function handler(req, res) {
   await connectMongo();
-  const { data } = req.query;
-  const username = data[0];
-  const url = data[1];
 
-  if (req.method != "PUT" || !username || !url) {
+  const { username, url } = req.query;
+
+  if (req.method != "GET" || !username || !url) {
     return res
       .status(400)
-      .json({ error: "Invalid request: 'username' and url 'required'" });
+      .json({ error: "Invalid request: username and link required" });
   }
 
-  const getProfile = await Profile.findOne({ username });
+  let getProfile = await Profile.findOne({ username });
 
   if (!getProfile) {
-    return res
-      .status(404)
-      .json({ error: "Invalid request: 'username' not found" });
+    try {
+      await Profile.create({
+        username,
+        views: 1,
+      });
+
+      getProfile = await Profile.findOne({ username });
+    } catch (e) {
+      const err = `ERROR creating profile: ${e}`;
+      console.log(err);
+      return res.status(500).json({ message: err });
+    }
   }
 
   const getLink = await Link.findOne({ username, url });
+
   if (getLink) {
     try {
       await Link.updateOne(
@@ -36,7 +48,9 @@
         }
       );
     } catch (e) {
-      console.log("ERROR incrementing link", e);
+      const err = `ERROR incrementing link stats: ${e}`;
+      console.log(err);
+      return res.res(500).json({ message: err });
     }
   }
 
@@ -48,6 +62,7 @@
         url,
         clicks: 1,
       });
+
       await Profile.updateOne(
         {
           username,
@@ -58,13 +73,18 @@
         { new: true, useFindAndModify: false }
       );
     } catch (e) {
-      console.log("ERROR creating link", e);
+      const err = `ERROR creating link: ${e}`;
+      console.log(err);
+      return res.res(500).json({ message: err });
     }
   }
 
   const date = new Date();
+
   date.setHours(1, 0, 0, 0);
+
   const getPlatformStats = await Stats.findOne({ date });
+
   if (getPlatformStats) {
     try {
       await Stats.updateOne(
@@ -76,7 +96,9 @@
         }
       );
     } catch (e) {
-      console.log("ERROR incrementing platform stats", e);
+      const err = `ERROR incrementing platform stats: ${e}`;
+      console.log(err);
+      return res.res(500).json({ message: err });
     }
   }
 
@@ -89,10 +111,11 @@
         users: 0,
       });
     } catch (e) {
-      console.log("ERROR creating platform stats", e);
+      const err = `ERROR incrementing platform stats: ${e}`;
+      console.log(err);
+      return res.res(500).json({ message: err });
     }
   }
 
-  const latestLink = await Link.findOne({ username, url });
-  res.status(201).json(latestLink);
+  return res.status(201).redirect(decodeURIComponent(url));
 }
```

Comment (eddiejaoude) [reply pages/api/users/[username]/links/[url].js:121]: Not sure I follow what the issue is, you shared the the same diff as I shared above it.

We don't need both files as they are almost identical, so we don't lose the history, move the file over to the new location with your changes on top - then we can discuss the changes

Comment (krshkun) [reply pages/api/users/[username]/links/[url].js:121]: @eddiejaoude when I move the files, it shows me 2 different changes, 1 for creating a new file and another for deleting

eddiejaoude:
Thanks Krish 👍 I will merge into a temporary branch and see what I can do

Issue #2523, opened 2022-12-28
[DOCS] Quickstart link accessibility


amandamartin-dev:
### Description

In the first 2 paragraphs of the linkfree quick start there are 2 links (SS below) that do not follow the same style and also are missing some accessibility features.

This issue should address the following:
Standardize link style 
Persistent underline on link (links should have a non-color indicator for accessibility)
Standard color/hover color for links (color indicator)
Standard focus style for links (keyboard only user)

This issue should also adjust any links in quickstart to open in a new tab so the user does not lose their place in the directions.  To make this accessible, warn the user in text that the link will open in a new tab.  

### Screenshots

![screenshot of linkfree quickstart with links in red squares](https://user-images.githubusercontent.com/97615019/209746783-b8918b26-52c5-4a4f-8758-c770050b165b.jpg)

neeraj2713:
I would like to work on this issue

neeraj2713:
Hey @amandamartin-dev , I would like to work on this issue :)

amandamartin-dev:
Hi @neeraj2713 I do not have the ability to assign issues but you are welcome to work on this. Thank you!

neeraj2713:
Alright!

PR #3276, opened 2023-01-12
Make links more accessible with custom link component (#2523)

clintonlunn:
## Changes proposed

I added a custom Link component that would give us a little bit better control over some of the accessibility concerns. I added tailwind classes to add text decoration and hover actions. I included these changes in both js and mdx files. 

I also fixed broken implementation of ClipboardCopy in [tags.mdx](https://github.com/EddieHubCommunity/LinkFree/blob/main/pages/docs/how-to-guides/tags.mdx)

## Screenshots

First screenshot to show without hover, second screenshot shows a link being hovered over (pointer invisible, sorry). 
image
image

## Note to reviewers

<a href="https://gitpod.io/#https://github.com/EddieHubCommunity/LinkFree/pull/3276"></a>


pages/docs/index.js
```diff
@@ -248,16 +249,9 @@ export default function DocsIndex() {
...
-            href="https://github.com/EddieHubCommunity/LinkFree/tree/main/pages/docs"
-            target="_blank"
-            rel="noreferrer"
-            className="text-sky-700 break-words"
-          >
-            https://github.com/EddieHubCommunity/LinkFree/tree/main/pages/docs
-          </a>
+          creating your Profile to more advanced topics. We welcome contributions, check out the&nbsp;
+          <Link url="https://github.com/EddieHubCommunity/LinkFree">LinkFree repo</Link> and the&nbsp;
+          <Link url="https://github.com/EddieHubCommunity/LinkFree/tree/main/pages/docs">documentation source</Link> on GitHub for more information.
```
eddiejaoude, line 254: I think this link has lost

```
target="_blank"
rel="noreferrer"
```

public/manifest.json
```diff
@@ -22,4 +22,4 @@
   "display": "standalone",
   "theme_color": "#000000",
   "background_color": "#ffffff"
-}
+}
```
eddiejaoude, line 25: Not sure this change is required?

components/Link.js
```diff
@@ -0,0 +1,15 @@
+import React from 'react'
+import NextLink from 'next/link'
+
+export default function Link({ url, children }) {
+    return (
+        <NextLink
+            href={url}
+            target="_blank"
```
eddiejaoude, line 8: I don't think every link should open on a new page

eddiejaoude (changes requested):
Thank you for your contribution 👍 

I have left some inline comments

public/manifest.json
clintonlunn: oops sorry, this was autoformatting

components/Link.js
clintonlunn: agreed, i can add a prop to allow for customization of the target

public/manifest.json
clintonlunn: resolved, my autoformatter removed the newline at the end of the file for some reason.

pages/docs/index.js
clintonlunn: resolved this, good catch!

clintonlunn:
@eddiejaoude Thanks for the review! I believe I resolved all of the issues you pointed out. 

However, I found some more occurrences of using next/link `Link` component that could be swapped out also (i.e. [Footer.js](https://github.com/EddieHubCommunity/LinkFree/blob/337cdecf982df319ded4f734d30d8681f26a0f13/components/Footer.js#L8) and some other places). If you agree with the direction of using a Link component wrapper around the next/link, do you want me to change the links out everywhere?

eddiejaoude:
> If you agree with the direction of using a Link component wrapper around the next/link, do you want me to change the links out everywhere?

Thank you for making the changes. I will review tomorrow (Sunday), but yes we want everywhere to be consistent, if you could do that, that would be great 👍

components/Link.js
```diff
@@ -0,0 +1,16 @@
+import React from 'react'
+import NextLink from 'next/link'
```
eddiejaoude, line 2: From the docs it is `import Link from 'next/link'`?

clintonlunn: if I `import Link from 'next/link' I get a parsing error 
image

I'm just creating an alias out of the Link to get rid of that parsing error. See docs: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/import#default_import

> Since the default export doesn't explicitly specify a name, you can give the identifier any name you like.

clintonlunn: I am exporting it as an alias to avoid a parsing error. 
image
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/import#default_import

components/Link.js
```diff
@@ -0,0 +1,14 @@
...
+
+export default function Link({ children, className, rel, ...restProps }) {
+    return (
+        <NextLink
+            rel={rel ? rel : "noreferrer"}
+            className={className ? className : "text-blue-600 underline decoration-dotted dark:text-blue-500 hover:underline hover:decoration-solid"}
+            {...restProps}
```
clintonlunn, line 9: rather than explicitly passing props into this component, I think it's a bit cleaner to just spread whichever props are not given default values down to the NextLink

components/Link.js
eddiejaoude: Ah I see, thank you for clarifying 👍 

Hmm maybe changing the name of our component would be more clear, great idea 💯 

Now what to call it? 🤔

clintonlunn: I'll throw out a few names, I'm unsure about what the name should be too. I almost like the approach of aliasing the next Link because a developer might search for a Link component, but maybe not a Custom/Hyper/Styled-Link.

CustomLink?
HyperLink?
StyledLink?

I'm definitely open to suggestions for names too.

eddiejaoude: `StyledLink` looks good to me 👍 - maybe let's go with this and we can always rename it again later on if someone can come up with a better name

eddiejaoude: What do you think?

eddiejaoude:
Thank you for your contribution. More conflicts are appearing, so I will merge to a temporary branch and try to resolve

clintonlunn:
@eddiejaoude apologies for the delay. I got caught up in work things. I made the changes locally and then started having issues bringing my fork up to date. I can resolve the merge conflicts on my end still if you want.

PR #3397, opened 2023-01-13
Avoid running reviewpad on automated pull requests

dan-mba:
## Changes proposed
Avoid running reviewpad on the 3 actions that automatically create PRs
1. Release
2. Testimonial
3. Formatting

<a href="https://gitpod.io/#https://github.com/EddieHubCommunity/LinkFree/pull/3397"></a>

### DIFF
--- a/.github/workflows/reviewpad.yml
+++ b/.github/workflows/reviewpad.yml
@@ -22,6 +22,12 @@ jobs:
   reviewpad_job:
     runs-on: ubuntu-latest
     name: Reviewpad
+    # Don't run reviewpad on automated PRs
+    if: |
+      (github.event_name != "pull_request_target") ||
+      ((github.event.pull_request.title != "Merge new release into main") &&
+      (github.event.pull_request.title != "Merge testimonial into main") &&
+      (github.event.pull_request.title != "Merge formatting into main"))
     steps:
       - name: Reviewpad
         uses: reviewpad/action@v3.x


Pradumnasaraf (commented):
Like this approach, as we have a standard naming for the auto-generated PRs.

Thanks, Dan

krshkun (approved):
Looks good. But why do we use ReviewPad with GitHub Actions? 🤔 https://docs.reviewpad.com/getting-started/installation

.github/workflows/reviewpad.yml
```diff
@@ -22,6 +22,12 @@ jobs:
   reviewpad_job:
     runs-on: ubuntu-latest
     name: Reviewpad
+    # Don't run reviewpad on automated PRs
+    if: |
+      (github.event_name != "pull_request_target") ||
+      ((github.event.pull_request.title != "Merge new release into main") &&
+      (github.event.pull_request.title != "Merge testimonial into main") &&
+      (github.event.pull_request.title != "Merge formatting into main"))
```
eddiejaoude, line 30: I wonder if we could prepend all the automated pull requests with something like `AUTOMATED: ...` then we can just check for that - this could come as a future improvement, it is not blocking this PR just was an idea that came to mind

eddiejaoude (approved):
Looks good 👍

.github/workflows/reviewpad.yml
dan-mba: I thinking changing 1 file instead of 4 is better in the short term. If this works as expected, I will come back to it.

PR #3431, opened 2023-01-14
data: khalidx3 

### LINKED ISSUE #3283: PlaceHolder text change
### Description

On the search page, the placeholder says `search user` but we also search on the basis of `tag/languages` 
image

we can change it to `search by name or tags` or we can keep it short as `search here` and add an example below it stating `search for name or tags ie. Manish or javascript`  if we consider this 
https://github.com/EddieHubCommunity/LinkFree/issues/3259

### Screenshots

_No response_

### Additional information

_No response_

khalidx3:
## Fixes Issue

Closes #3283 
## Changes proposed

To change Search bar PlaceHolder from Search Users to Search Users/tags/languages/etc

## Screenshots

ssry

## Note to reviewers

<a href="https://gitpod.io/#https://github.com/EddieHubCommunity/LinkFree/pull/3431"></a>

### DIFF
--- a/data/khalidx3.json
+++ b/data/khalidx3.json
@@ -1,0 +1,25 @@
+{
+  "name": "Khalid Hasan",
+  "displayStatsPublic": true,
+  "type": "personal",
+  "bio": "Fitness Tech and Books thats what describes me and yeah I'm an engineer",
+  "avatar": "https://github.com/khalidx3.png",
+
+"links": [
+    {
+      "name": "Follow me on Twitter",
+      "url": "https://twitter.com/khalidansariX3",
+      "icon": "FaTwitter"
+    }
+  {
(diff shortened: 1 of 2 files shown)


data/khalidx3.json
```diff
@@ -0,0 +1,25 @@
+{
```
ManishBisht777, line 1: Pr for Adding your profile already open so you can remove it from here 

I think you actually made changes in the same fork as profile addition that's why there are same changes in both of your pr

khalidx3: Alright got it that's my bad will do it.

EmmaDawsonDev (changes requested):
Your profile looks good. If you can remove the other file from this PR then your profile can be merged.

PR #3496, opened 2023-01-15
[FIX] fixes the links view count not being sorted

### LINKED ISSUE #3434: Views Order Not correct
### Description

On toggling the sorting order of `Links` in profile page the views does not sort

### Default

image

### On sort the order changes but the view does show in the same

image
 

### Screenshots

_No response_

### Additional information

_No response_

Aadarsh805:
## Fixes Issue

closes #3434 

## Changes proposed
This was happening due to not having the correct key property while mapping over the link

Instead of using `index` as key, changed it to `link.name`
the Link has `name, icon and url` values in it, the name seemed the most optimal

- [x ] My code follows the code style of this project.
- [ x] All new and existing tests passed.
- [ x] This PR does not contain plagiarized content.
- [ x] The title of my pull request is a short description of the requested changes.

## Screenshots

Screenshot from 2023-01-15 22-01-45

image

## Note to reviewers

If having `link.name` as the key is not the appropriate option, we can add an `id` on each link before mapping over it like this
```
  data.links = data.links.map((link, i) => ({ id: i, ...link }));
```
By this, each link will also have a unique `id` field which can be used as a key

<a href="https://gitpod.io/#https://github.com/EddieHubCommunity/LinkFree/pull/3496"></a>

### DIFF
--- a/components/user/UserLinks.js
+++ b/components/user/UserLinks.js
@@ -8,10 +8,10 @@ export default function UserLinks({ BASE_URL, data }) {
       {data.links && (
         <div className="flex flex-col items-center w-full">
           {data.links &&
-            data.links.map((link, index) => (
+            data.links.map((link) => (
               <UserLink
                 BASE_URL={BASE_URL}
-                key={index}
+                key={link.name}
                 link={link}
                 username={data.username}
                 displayStatsPublic={data.displayStatsPublic}
--- a/components/user/UserLinks.js
+++ b/components/user/UserLinks.js
@@ -2,6 +2,8 @@ import UserLink from "./UserLink";
 import Alert from "../Alert";
 
 export default function UserLinks({ BASE_URL, data }) {
+  data.links = data.links.map((link, i) => ({ id: i, ...link }));
+
   return (
     <>
       {!data.links && <Alert type="info" message="No links found" />}
@@ -11,7 +13,7 @@ export default function UserLinks({ BASE_URL, data }) {
             data.links.map((link) => (
               <UserLink
                 BASE_URL={BASE_URL}
-                key={link.name}
+                key={link.id}
                 link={link}
                 username={data.username}
                 displayStatsPublic={data.displayStatsPublic}


Aadarsh805:
@ChinmayMhatre can you have a look at this
Also tell me if using the name as key is a good practice or not, or I'll switch to the other method to fix the issue as I mentioned above

ChinmayMhatre:
> @ChinmayMhatre can you have a look at this
> Also tell me if using the name as key is a good practice or not, or I'll switch to the other method to fix the issue as I mentioned above 

I think using the map's index would be better than using the name.

ChinmayMhatre:
> @ChinmayMhatre can you have a look at this
> Also tell me if using the name as key is a good practice or not, or I'll switch to the other method to fix the issue as I mentioned above 

I'll take a look at the code at night.

Aadarsh805:
> > @ChinmayMhatre can you have a look at this
> > Also tell me if using the name as key is a good practice or not, or I'll switch to the other method to fix the issue as I mentioned above
> 
> I think using the map's index would be better than using the name.

Using index was the reason it was not working

ChinmayMhatre:
> > > @ChinmayMhatre can you have a look at this
> > > Also tell me if using the name as key is a good practice or not, or I'll switch to the other method to fix the issue as I mentioned above
> > 
> > I think using the map's index would be better than using the name.
> 
> Using index was the reason it was not working

I meant this

Note to reviewers

If having link.name as the key is not the appropriate option, we can add an id on each link before mapping over it like this

data.links = data.links.map((link, i) => ({ id: i, ...link })); 

By this, each link will also have a unique id field which can be used as a key

You could try using this id

ManishBisht777 (commented):
Looks good to me

eddiejaoude:
What about using `url` that will be unique?

Aadarsh805:
> url

can't say, it's user-defined that's why

Aadarsh805:
> What about using `url` that will be unique?

can't say, it's user-defined that's why

eddiejaoude (approved):
The links do have `id`s from the DB, but we can merge this now to fix it, and then improve further in a new PR

Thank you all for the contribution and collaboration 👍

PR #3584, opened 2023-01-17
Vishrut19 patch 1

Vishrut19:


ChinmayMhatre (changes requested):
Please revert the package-lock.json from the commit . To do this you can navigate to your branch and run `git checkout main -- package-lock.json`.

I case you need more help you can refer this video by Eddie https://www.youtube.com/watch?v=YUHS5Ana7Jw

PR #3632, opened 2023-01-18
Storybook v7

dan-mba:
## Changes proposed

Re-enable storybook

<a href="https://gitpod.io/#https://github.com/EddieHubCommunity/LinkFree/pull/3632"></a>


config/logger.js
```diff
@@ -9,8 +9,18 @@ const config = {
...
-const logger = PinoLogger.pino(
-  process.env.NODE_ENV === "development" ? config : {}
-);
+let logger
+
+if (process.env.STORYBOOK_RUN) {
+  logger = {
+    child: console.log,
+    info: console.log,
+    error: console.log,
```
eddiejaoude, line 18: By default I think Pino it goes to standard out, would this still be needed?

eddiejaoude: Ah interesting, I selected Pino because it is meant to work in the browser over something like Winston - ok no problem, we can always tweak in the future

eddiejaoude (dismissed):
Looks good 👍

config/logger.js
dan-mba: According to the [Pino Docs](https://github.com/pinojs/pino/blob/master/docs/browser.md), in order for it to work with the browser it needs to be used with [Browserify](https://github.com/browserify/browserify).
Browserify is a pre-processer that resolves all the requires & polyfills node globals (process, __filename, __dirname, etc).
This won't work easily with the Storybook setup, so the polyfill I used seems like a simpler solution.

BTW, the logger code never runs, because it is called from `getServerSideProps()` which isn't run by Storybook.
The component is rendered with the props specified in the story.

dan-mba:
@eddiejaoude Does LinkFree use GitHub Pages?

If not, I think deploying Storybook to GitHub Pages might be useful so people can see what their profile will look like before a PR gets merged.

eddiejaoude:
> Does LinkFree use GitHub Pages?
> If not, I think deploying Storybook to GitHub Pages might be useful so people can see what their profile will look like before a PR gets merged.

We don't use GitHub Pages at the moment. I think that is a great idea 👍

config/logger.js
eddiejaoude: ok thank you for the detailed reply Dan 👍

.github/workflows/storybook.yml
```diff
@@ -8,9 +13,10 @@ jobs:
...
-          node-version: "16"
+          node-version: "18"
           cache: "npm"
+
       - name: install dependencies
         run: npm ci
-      # - name: run build
-      #   run: npm run build-storybook
+      - name: run build
+        run: npm run build-storybook
```
eddiejaoude, line 22: Could both Actions be merged into one and have the deploy with an `if` condition for the branch `main`?

eddiejaoude (dismissed):
Looks good 👍 

Just 1 comment about possibly combining the GitHub Actions

.github/workflows/storybook.yml
eddiejaoude: It is not a blocker for this PR, it was just an idea. For example

Action
- build
- deploy with the `if` condition, I think something like this `if: github.ref == 'refs/heads/master'`

dan-mba: I got the actions combined now.

And the ifs work correctly

.github/workflows/storybook.yml
```diff
@@ -8,9 +20,33 @@ jobs:
...
+        with:
+          path: './storybook-static'
+
+# Deployment job
+  deploy:
+    environment:
+      name: github-pages
+      url: ${{ steps.deployment.outputs.page_url }}
+    runs-on: ubuntu-latest
+    if: github.event_name == 'push'
```
eddiejaoude, line 46: I think this will run for every push, where I think you wanted it just for changes in `main`? Maybe something like this `refs/heads/main`?

dan-mba: the `on` settings are made so it will only run on main.
```yml
push:
    branches:
      - main
    paths:
      - pages/**
      - components/**
      - stories/**
```

eddiejaoude: But it has `pull_request` on there also? https://github.com/EddieHubCommunity/LinkFree/pull/3632/files#diff-97cca8fb169bc3951469886bf2f73b7bcc362a6b6da11325c7a617f69a797248R10

eddiejaoude: I am just thinking for `pull_request` it should only build and not deploy?

eddiejaoude (approved):
Looks good 👍

Issue #900, opened 2022-01-03
[FEATURE] Use search engine for Search


developer-diganta:
### What would you like to share?

I was thinking if we can implement [FuseJs](https://fusejs.io/) for search?

emmalearnscode:
Can you explain a bit more about it and how it could benefit the project

developer-diganta:
@emmalearnscode I am not sure if this will be of some huge help but I thinkFuseJs can help get close search results. For example, if someone's name is Jenny and someone types just Jeny then FuseJs can list out the name (it does it on the closeness of the word entered by the user).

ManishBisht777:
i want to work on this

SaraJaoude:
Closing as this is a duplicate of #8169

PR #3721, opened 2023-01-20
[FEATURE] fixed tags duplicate issue in discovery page due to case sensitivity

### LINKED ISSUE #3604: [FEATURE] tags in discovery duplicate due to case sensitivity
### Description

The tags section in the discovery tab has duplicate entries du to case sensitivity.
I think all tags should be not case sensitiv in my opinion.

### Screenshots

image

### Additional information

_No response_

rwiteshbera:
Closes #3604

## Fixes Issue
Fixed tags duplicate issues in discovery page due to case sensitivity

Before: `JavaScript(10), javascript(5), Javascript(2)`
After: `javascript(17)`

## Changes proposed

While displaying the tags, I have added a snippet that will do the following task with all the tags:
1. Trim whitespace from both sides of the string.
2. Converted into lowercase
3. Replaced all the special characters with ''. (Example: react.js -> reactjs)
4. If the current tag name already exists as a key in the accumulated object, increment the value by the current tag's total count. If the tag name does not exist in the object, add a new key with the current tag's total count as the value.

## Screenshots

image

## Note to reviewers
Please let me know if there are any changes or updates needed before it can be merged. I am open to suggestions and feedback to ensure that the code is of the highest quality.

<a href="https://gitpod.io/#https://github.com/EddieHubCommunity/LinkFree/pull/3721"></a>


eddiejaoude (changes requested):
This is a great idea, but I think it would be best to happen on the server side here https://github.com/EddieHubCommunity/LinkFree/blob/main/pages/api/discover/tags.js

I think to keep things super simple we could lowercase all tags

ManishBisht777 (commented):
on the backend part stats are still set differently `reactjs`, and `react js` so if I search for `reactjs` I will not get results for `react js` 

so I think putting the data in the server in the right format seems  a better option

rwiteshbera:
Hi, @eddiejaoude, I guess it would be beneficial to reorganize and reformat the current data by converting it to a standardized format, such as making all letters lowercase and removing any special characters or spaces.

eddiejaoude:
I think then we are trying to build a search engine, which we don't want to do. There is an issue to improve the search using libraries/tools that focus on search https://github.com/EddieHubCommunity/LinkFree/issues/900

eddiejaoude:
Thank @rwiteshbera for the improvement, the tags do looks better when listed on the Discover page, howeer then the results don't match when a tag is clicked and searched. 

This could be solved when performing the search applying the same logic to the users tag, but then we would need to apply the same for the users profile tag list also.

rwiteshbera:
Could you please provide me with some screenshots so I can proceed with my local setup, as it is currently functioning properly? In my case, both results look the same when I am searching or clicking on the user's tag.

eddiejaoude:
How does this affect the search? For example instead of `JavaScript(10)`, `javascript(5)`, `Javascript(2)` it shows `javascript(17)` but when searching will it show the correct results of 17 or something else?

rwiteshbera:
Hi, @eddiejaoude , I am getting correct results here. 
Screenshot 2023-01-27 164335

image

eddiejaoude:
Thank you for your contribution. This recently got removed from the Discovery page. So I will close this PR.

We are creating more issues, so keep an eye on there for more

Issue #2600, opened 2023-01-02
[FEATURE] Group events together


eddiejaoude:
### Description

On the community events page, we have duplicate events, but we can group them together and show multiple people attending them

I think stacking the avatars would look good 

<img width="218" alt="Screenshot 2023-01-02 at 22 01 21" src="https://user-images.githubusercontent.com/624760/210280761-a42de8c4-bf6e-450b-98ce-edc5b42b4a80.png">

https://tailwindui.com/components/application-ui/elements/avatars#component-00ccf3da2333eae32886bbc1d1ee362c

ThorOnTheRocks:
@eddiejaoude Can I get this one?

eddiejaoude:
Sure, assigned to you 👍

eddiejaoude:
I am not sure what would be best to group them on, maybe the url?

thebarshablog:
How far are you on this @ThorOnTheRocks? Do you mind an extra pair of hands on this?
It would be lovely to know you and work with you!

ThorOnTheRocks:
@eddiejaoude I guess grouping them on the url would be a good choice, showing only one card if it is the same event and the stacked avatar on the right corner. Regarding the UI I was already thinking about the stacked avatar from tailwind, it's the way to go I think.

@thebarshablog I haven't even started to be honest, my plan was to start later today after work but yeah that would be amazing actually to work together! We should split the tasks I guess like someone taking care of the UI and another about the logic perhaps? Just to avoid working on the same files maybe...I open to any suggestions really so what you think it's best!

thebarshablog:
Sounds great! @ThorOnTheRocks. Are you open to getting over a GMeet or otherwise to speak elaboratively on it? Let me know, I'll make arrangements

eddiejaoude:
> I was already thinking about the stacked avatar from tailwind, it's the way to go I think.

I have paid for the Tailwind components, so I can get you the code for that component to use on the project

```js
export default function Example() {
  return (
    <>
      <div className="isolate flex -space-x-1 overflow-hidden">
        <img
          className="relative z-30 inline-block h-6 w-6 rounded-full ring-2 ring-white"
          src="https://images.unsplash.com/photo-1491528323818-fdd1faba62cc?ixlib=rb-1.2.1&ixid=eyJhcHBfaWQiOjEyMDd9&auto=format&fit=facearea&facepad=2&w=256&h=256&q=80"
          alt=""
        />
        <img
[...]

PR #3841, opened 2023-01-23
Attending/Speaking information added in personal event card

### LINKED ISSUE #3711: [FEATURE] Events - add speaking/attending details to events
### Description

**EDIT:** This issue will only handle adding speaker details to personal profile pages.  A new issue can be opened later for the community page after #2600 is completed

Relates to #2600 

On the community events it would be nice to know which events folks are attending vs speaking so that interested folks could connect with them or discover sessions they may want to attend.  Proposal to add this as an option to events.  To keep the cards from getting too big, this could be an expandable option in the UI (open to other suggestions)

**Changes proposed:**
add "isSpeaking" or "isAttending" as a required option to events.  This will allow other folks heading to these events to connect with friends

if "isSpeaking", then a required field is "session" which would contain session details or title (there should be a limit on the length)

**Challenges:**  Due to the 1:many nature of events this may best be discussed after events are grouped in #2600 to figure out the best way to manage this in the UI on the community page.  Personal profiles will be easier in this respect.

### Screenshots

_No response_

### Additional information

_No response_

MitAbhay:
## Fixes Issue
closes #3711 

## Screenshots

image

## Note to reviewers

<a href="https://gitpod.io/#https://github.com/EddieHubCommunity/LinkFree/pull/3841"></a>


components/user/UserEvents.js
```diff
@@ -5,6 +5,7 @@ import Alert from "../Alert";
 import EventKey from "../event/EventKey";
 
 export default function UserEvents({ data }) {
+  console.log(data);
```
ManishBisht777, line 8: You can remove this

components/event/EventCard.js
```diff
@@ -35,6 +35,26 @@ export default function EventCard({ event, username }) {
             )}
         </div>
         <div className="flex-1 space-y-1">
+          <div className="flex space-x-2">
+            <p>
```
ManishBisht777, line 39: We can add a condition in case the user forgot to mention the fields 
image

pages/docs/how-to-guides/events.mdx
```diff
@@ -24,6 +24,10 @@ _If you need help on how to edit this file, please see the [Editing guide](/docs
 <ClipboardCopy>
 ```js
 {
+  "userStatus": "speaking",
```
ManishBisht777, line 27: Maybe Add a separate pr for changes in docs? @eddiejaoude

stevenovak123 (dismissed):
LGTM

components/event/EventCard.js
```diff
@@ -9,7 +9,7 @@ import { ReactMarkdown } from "react-markdown/lib/react-markdown";
 
 import FallbackImage from "../FallbackImage";
 
-export default function EventCard({ event, username }) {
+export default function EventCard({ event, username, name }) {
```
ManishBisht777, line 12: there is no need of this `name` as name is already present in `userName`

components/event/EventCard.js
```diff
@@ -35,6 +35,28 @@ export default function EventCard({ event, username }) {
             )}
         </div>
         <div className="flex-1 space-y-1">
+          {!username && event.userStatus && (
```
ManishBisht777, line 38: It should be `username && event.userStatus` as `userName` will always be there

or we can even remove userName and keep it as just `event.userStatus`

components/event/EventCard.js
```diff
@@ -35,6 +35,28 @@ export default function EventCard({ event, username }) {
...
+                  style={{
+                    color: event.color,
+                  }}
+                >
+                  {event.userStatus}
+                </span>{" "}
+                {event.userStatus == "speaking" && "at"} this event.
+              </p>
+              {event.userStatus == "speaking" && (
+                <p>
```
ManishBisht777, line 53: There is a need of condition here as well as `speakingTopicDetails` is also needed

`{event.userStatus == "speaking" && event.speakingTopicDetails && (
                 <p>
                   <span className="font-bold">Topic: </span>{" "}
                   {event.speakingTopicDetails.topic} 
                 </p>
              )}`

ManishBisht777 (commented):
I left some inline comments

components/event/EventCard.js
MitAbhay: We want to show the status only on the personal event card, not on the community event card. That's why this condition is there.

ManishBisht777: i think its good to have it on both places but not sure @eddiejaoude any suggestions

components/event/EventCard.js
MitAbhay: Topic details are not necessary I think.

components/event/EventCard.js
MitAbhay: On the community events card, we are showing all the attendees/speakers on a single card, so this will not work then. Look at this issue #2600

components/event/EventCard.js
ManishBisht777: yeah, but according to the code if the `SpeakingTopicDetails` is not present then it just shows `Topic :`

components/event/EventCard.js
ManishBisht777: Oh if that's the case maybe we can just pass a flag here? along with userName as in the props both `userName` and `name` are present which are technically the same but coming from diff pages

`<EventCard username={data.name} event={event} key={index} isUserPage={true} />`

and in the props

`export default function EventCard({ event, username, isUserPage })`

components/event/EventCard.js
MitAbhay: Ohk I got it

components/event/EventCard.js
MitAbhay: ok

pages/docs/how-to-guides/events.mdx
eddiejaoude: Ideally code changes and docs and tests should all be in the same PR if related

ManishBisht777 (commented):
Looks good to me :)

pages/docs/how-to-guides/events.mdx
```diff
@@ -25,6 +25,10 @@ _If you need help on how to edit this file, please see the <Link href="/docs/how
...
+  "speakingTopicDetails": {
```
ManishBisht777, line 29: is there any other fields inside `speakingTopicDetails`? if not just `topic` filed would be good instead of an object

MitAbhay: Maybe we can add other fields later, that's why I made it as an object.  Like `"description" : ""  etc.`

eddiejaoude:
Great collaboration everyone 👍

PR #3894, opened 2023-01-24
Add workflow dispatch to prettier action

dan-mba:
Add workflow dispatch to the prettier action.
The need for this is merging from main is causing misc profiles to be included in PRs.


eddiejaoude (approved):
Looks good 👍

PR #3922, opened 2023-01-24
Cleaner event listener function call

Ananya2001-an:

## Note to reviewers

Since no arguments are being passed to the fetchRandom function it’s better to call it directly rather than creating a callback.

<a href="https://gitpod.io/#https://github.com/EddieHubCommunity/LinkFree/pull/3922"></a>

### DIFF
--- a/pages/discover.js
+++ b/pages/discover.js
@@ -96,7 +96,7 @@ export default function Popular({ data, BASE_URL }) {
             </h2>
             <button
               className="inline-flex gap-3 items-center justify-center rounded-full border border-transparent bg-white px-2 py-1 text-base font-medium text-indigo-600 border-indigo-600 hover:bg-indigo-50"
-              onClick={() => fetchRandom()}
+              onClick={fetchRandom}
               aria-label="refresh random profiles"
             >
               <TbRefresh />


pages/discover.js
```diff
@@ -96,7 +96,7 @@ export default function Popular({ data, BASE_URL }) {
             </h2>
             <button
               className="inline-flex gap-3 items-center justify-center rounded-full border border-transparent bg-white px-2 py-1 text-base font-medium text-indigo-600 border-indigo-600 hover:bg-indigo-50"
-              onClick={() => fetchRandom()}
```
ManishBisht777, line 99: It make the `fetchRandom` function run when the component mount itself refreshing the profiles 2 times first the onclick and 2nd from `getServerSideProps`

Ananya2001-an: @ManishBisht777 Okay, I didn't know about it earlier. Learnt something new. 
So basically if I just write the function as it is it will be called directly no matter what even if the onClick event hasn't taken place, right ?

ManishBisht777: > @ManishBisht777 Okay, I didn't know about it earlier. Learnt something new. So basically if I just write the function as it is it will be called directly no matter what even if the onClick event hasn't taken place, right ?

yep

PR #3961, opened 2023-01-25
feat: group links #2437

### LINKED ISSUE #2437: [FEATURE] Group links together
### Description

Sometimes we want to group links together.

Maybe all links can fall in `default` but then people could add other groups like `community` or `business` etc

I think this could be done with a group name property in the links section and if not set use `default`

```js
"links": [
   {
      ...
      "group": "business",
      ...
   }
]
```

**OR**

Have a field in the link object for the group, for example

```js
"links": [
   {
      "group": "Socials",
      ...
   }
]
```

Similar to @madzadev's profile on BioLink https://bio.link/madza

Screenshot 2023-01-22 at 17 26 33

### Screenshots

_No response_

### Additional information

_No response_

eddiejaoude:
## Fixes Issue

fixes #2437 

NO GROUPs

Screenshot 2023-01-25 at 10 10 33

WITH GROUPs

Screenshot 2023-01-25 at 10 10 20

### DIFF
--- a/components/user/UserLinks.js
+++ b/components/user/UserLinks.js
@@ -2,7 +2,14 @@ import UserLink from "./UserLink";
 import Alert from "../Alert";
 
 export default function UserLinks({ BASE_URL, data }) {
+  const defaultBucket = "Others";
   data.links = data.links.map((link, i) => ({ id: i, ...link }));
+  const buckets = data.links.reduce((acc, obj) => {
+    const group = obj.group || defaultBucket;
+    const curGroup = acc[group] ?? [];
+
+    return { ...acc, [group]: [...curGroup, obj] };
+  }, {});
 
   return (
     <>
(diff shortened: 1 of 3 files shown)


ChinmayMhatre:
There is not Gitpod button for this PR 🤔

eddiejaoude:
Oh that is strange. It happens sometimes. I still think it can be opened in Gitpod by prepending their url

ChinmayMhatre:
This looks Great! I was wondering since there are around 1k users with profiles already created, could we actually check if there are any groups and skip the groups entirely if none are present instead of adding them to `others` ?

ChinmayMhatre:
> Oh that is strange. It happens sometimes. I still think it can be opened in Gitpod by prepending their url

Yes! Just checked. This looks great!

eddiejaoude:
Thank you for reviewing @ChinmayMhatre 👍

PR #4035, opened 2023-01-26
feat: start calls at same time

ManishBisht777:
## Changes proposed
In the current code and there are `4 calls being made` and all of them `don't depend on each other` so rather than waiting for one call to finish then start calling next what we can do is `start calling all 4 of them at the same time` and wait for them later

## Note to reviewers

It does seem to work on paper but I didn't see any performance changes when checked on the lighthouse. let me know your thoughts.

<a href="https://gitpod.io/#https://github.com/EddieHubCommunity/LinkFree/pull/4035"></a>


pages/discover.js
```diff
@@ -11,40 +11,43 @@ export async function getServerSideProps(context) {
...
+    `${process.env.NEXT_PUBLIC_BASE_URL}/api/discover/trending`
+  );
+
   try {
-    const res = await fetch(
-      `${process.env.NEXT_PUBLIC_BASE_URL}/api/discover/popular`
-    );
-    data.popular = await res.json();
-  } catch (e) {
+    data.popular = await (await popularPromise).json()
```
eddiejaoude, line 29: I think a `promise all` would be more efficient otherwise this one is still blocking the remaining waits in this order

ManishBisht777: The calls have been initiated so when we `wait for one` the rest of them are either processed or running in the background

the eg:- we are waiting for the `popularPromise` but other 3 promises are running so when we finish with the `popularPromise` rest of them are either finished or about to finish so 

rest of the `await (await Promise).json()` will not taking any time or take a very less time 

Plus the `Promise.all` is all or nothing so if we end up `rejecting one promise` the whole list of promises reject hence getting an `empty data as prop`

eddiejaoude: All requests are to the same server, it is unlikely 1 will throw an exception. If we are trying to improve the performance, we should use promise all to maximise the performance and truly run them in parallel

ManishBisht777: ahh, Got it

eddiejaoude:
No problem. Thank you for making the changes, they look good 👍 

Unfortunately, since then the Discover page has been removed as people were gaming the system. So I will close this PR. Please have a look at open issues that are unassigned and have the label `ready for dev`

PR #4091, opened 2023-01-27
feat: basic login with github

eddiejaoude:
- [x] statistics page using https://recharts.org/en-US/api
- [x] login with GitHub
- [ ] ~~feature flag~~
---
move to another PR
- [ ] save session to db
---
- [x] save daily link clicks per link (already merged)

Screenshot 2023-01-28 at 19 04 33

<a href="https://gitpod.io/#https://github.com/EddieHubCommunity/LinkFree/pull/4091"></a>


pages/api/auth/[...nextauth].js
```diff
@@ -0,0 +1,22 @@
...
+export const authOptions = {
+  providers: [
+    GithubProvider({
+      clientId: process.env.GITHUB_ID,
+      clientSecret: process.env.GITHUB_SECRET,
+      profile(profile) {
+        return {
+          id: profile.id.toString(),
+          name: profile.name ?? profile.login,
+          username: profile.login,
```
Comment (eddiejaoude) [inline pages/api/auth/[...nextauth].js:13]: bug: `username` (`profile.login`) is not coming through?

pages/account/statistics.js
```diff
@@ -0,0 +1,83 @@
...
+        destination: "/",
+        permanent: false,
+      },
+    };
+  }
+
+  let data = {};
+  try {
+    const res = await fetch(
+      `${process.env.NEXT_PUBLIC_BASE_URL}/api/account/statistics`,
```
eddiejaoude, line 26: @Cahllagerfeld this is the request that fails, the api rejects with no session found - but visit the API path in the browser is fine

iamthecloverly (dismissed):

pages/account/statistics.js
eddiejaoude: Resolved by passing 

```js
headers: {
   cookie: context.req.headers.cookie || "",
}
```

pages/account/statistics.js
```diff
@@ -77,7 +79,9 @@ export default function Search({ data, profile }) {
   const dateTimeStyle = {
     dateStyle: "short",
   };
-  const dailyViews = data.profile.daily.slice(-30).map((day) => {
+  
+  let sortedByDate = data.profile.daily.sort((a, b) => new Date(a.date) - new Date(b.date));
```
eddiejaoude, line 82: I think we can sort on the server side [here](https://github.com/EddieHubCommunity/LinkFree/blob/auth-stats/pages/api/account/statistics.js#L36) (but I sorted by views instead of date, I think I did a copy and paste and forgot to update it 🤦‍♂️ )

eddiejaoude: Thank you for noticing my mistake 👍 , I just pushed a fix

Pradumnasaraf (commented):
Looks good!

PR #4272, opened 2023-01-31
data: Chrisidoko

Chrisidoko:


rahulpandey70 (dismissed):
LGTM

Panquesito7 (approved):
Thanks! 🎉

data/Chrisidoko.json
```diff
@@ -0,0 +1,50 @@
+{
+   "name":"Eddie Jaoude",
```
eddiejaoude, line 2: I don't think this is correct?

eddiejaoude (changes requested):
Incorrect name

data/Chrisidoko.json
Chrisidoko: Yes i am aware i will make corrections to it

Chrisidoko: Done

Panquesito7 (approved):
Thanks! 🎉

kumarsonsoff3 (approved):
Looks Good, Thanks! 🎉

PR #4030, opened 2023-01-26
Updated Search page UI

Ananya2001-an:
## Fixes Issue

closes #3849 

## Changes proposed

Added filtering option with tags on Search page as well. Users can select and deselect tags to get the desired results based on the combination chosen. For example: if I select tags open source and DevOps then only those users that have both of these tags will be shown. You can click on the tag again to remove the selection. You can also manually type out in the search bar to get the results otherwise use tags directly. When the user selects a tag from the discover page or their LinkFree profile, it will redirect them to the search page showing the results along with the highlighted tag.

## Check List (Check all the applicable boxes) 

- [x] My code follows the code style of this project.
- [x] My change requires changes to the documentation.
- [ ] I have updated the documentation accordingly.
- [ ] All new and existing tests passed.
- [x] This PR does not contain plagiarized content.
- [x] The title of my pull request is a short description of the requested changes.

## Screenshots

image

image

## Note to reviewers

<a href="https://gitpod.io/#https://github.com/EddieHubCommunity/LinkFree/pull/4030"></a>

PR #4375, opened 2023-02-02
Multi tag search

eddiejaoude:
## Fixes Issue

Minor tweaks from the great work by @Ananya2001-an on https://github.com/EddieHubCommunity/LinkFree/pull/4030

## Note to reviewers

<a href="https://gitpod.io/#https://github.com/EddieHubCommunity/LinkFree/pull/4375"></a>

### DIFF
--- a/components/Tag.js
+++ b/components/Tag.js
@@ -1,17 +1,33 @@
 import Link from "../components/Link";
 import { abbreviateNumber } from "../services/utils/abbreviateNumbers";
 
-export default function Tag({ name, total }) {
+export default function Tag({ name, total, path, currentInput, selected, method }) {
+
   return (
     <div className="relative">
       <Link
-        href={`/search?search=${name}`}
-        className="flex flex-row p-1 m-2 rounded-lg text-sm font-mono border-2 cursor-pointer shadow-none hover:border-orange-600"
+        onClick={()=>{
+          if(!selected){
+            if(currentInput !== '' && currentInput !== undefined)
+            {
+              method(`${currentInput.toLowerCase().split(',').filter(input => input !== '').join(',')},${name}`)
+            }
+            else if(currentInput !== undefined){
+              method(`${name}`)
+            }
+            else{}
+          }else{
+            method(currentInput.toLowerCase().split(',').filter(input => input !== '' && input !== name.toLowerCase()).join(','))
+          }
+        }}
+        href={path && path ==='/search'? '':`/search?search=${name}`}
+
+        className={`flex flex-row p-1 m-2 rounded-lg text-sm font-mono border-2 cursor-pointer shadow-none ${selected? 'hover:border-white bg-orange-600 text-white' : 'hover:border-orange-600'}`}
       >
         {name}
       </Link>
       {total && (
-        <div className="absolute inline-block top-0 right-0 bottom-auto left-auto translate-x-1/4 -translate-y-1/3 rotate-0 skew-x-0 skew-y-0 scale-x-100 scale-y-100 py-1 px-1.5 text-xs leading-none text-center whitespace-nowrap align-baseline font-bold bg-orange-600 text-black rounded-full z-10">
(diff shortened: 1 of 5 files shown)


Ananya2001-an:
@eddiejaoude Hey Eddie. I will review the code in 1 hour or so. I have got some urgent work right now. Sorry for the delay but I will definitely check everything and tell you if it's working fine or not.

eddiejaoude:
No problem @Ananya2001-an , please do not stress, I look forward to hearing from you when you have some time

components/Tag.js
```diff
@@ -1,17 +1,62 @@
...
+  selected,
+  setInputValue,
+}) {
   return (
     <div className="relative">
       <Link
-        href={`/search?search=${name}`}
-        className="flex flex-row p-1 m-2 rounded-lg text-sm font-mono border-2 cursor-pointer shadow-none hover:border-orange-600"
+        onClick={() => {
+          if (!currentInput) {
```
Ananya2001-an, line 16:
```suggestion
          if (currentInput === undefined) {
```

components/Tag.js
```diff
@@ -1,17 +1,62 @@
...
+            return;
+          }
+          if (!selected) {
```
Ananya2001-an, line 19: Replace the whole part with this please. 

```
if (!selected) {
            if (currentInput !== "") {
              setInputValue(
                `${currentInput
                  .toLowerCase()
                  .split(",")
                  .filter((input) => input !== "")
                  .join(",")},${name}`
              );
            }

            else {
              setInputValue(`${name}`);
            }
          }

          else{
            setInputValue(
              currentInput
                .toLowerCase()
                .split(",")
                .filter((input) => input !== "" && input !== name.toLowerCase())
                .join(",")
            );
          }
```

Ananya2001-an: @eddiejaoude I checked the code and the tags on the search page were not getting enabled directly. So I tweaked a bit and then it worked fine. Please make the above 2 changes that I have asked.

eddiejaoude: Thank you @Ananya2001-an for being super productive with this feature. I tweaked it further to simplify the code as it was quiet complicated, you had almost built a search engine 😂 

We will add a search engine later, so I was keen to keep this simple, have a look and let me know what you think (java and javascript are still both triggered, but I was thinking, where do we stop, there are many other namings that will also clash and I don't want to manage a whole dictionary of endless possibilities)

I look forward to hearing your feedback, thanks again

Ananya2001-an: @eddiejaoude I really did overthink everything and actually created a search engine 😂
But I agree with you on the fact to keep it simple. I tested it out right now and it works great just a few things I want you to change if that's possible:

1. In the selected condition make the tag name lowercase so that it gets highlighted no matter the input otherwise user has to keep in mind the exact case.

2. Remove tagNames array in the search.js file it's no longer being used

3. If possible for the selected condition you can convert the inputValue into an array with all values in lowercase and then check if the tag.name.toLowercase() is present or not. This will remove the error of java and react getting highlighted when clicking on javascript and reactjs respectively.

4. One last thing that on the user profile the links are now blue in colour you need to adjust the css there.

That's it I guess. Overall I am really happy with the simplified working ❤️

Ananya2001-an: Also it was right to remove the tags from the discover page. It was becoming quite redundant.

eddiejaoude: @Ananya2001-an great points 💯 

1. ✅ Good point
2. ✅ Thank you, good spot
3. Sounds like a great idea, but not sure I follow how to do it. Could you create a new PR for this?
4. ✅ Good spot

Ananya2001-an: Sure I will raise a new PR to fix the 3rd issue. Besides that congrats on the new feature 👏

Ananya2001-an: Also in the same PR I will make a change for the Home page where we need to update the Search page screenshot

eddiejaoude: Great 👍  Thank you 🚀

PR #4440, opened 2023-02-03
Reduce _app bundle size to improve load time

dan-mba:
## Changes proposed
Import IconContext directly in NavBar.
Use getIcon to import icons in NavBar & Footer.

### DIFF
--- a/components/navbar/Navbar.js
+++ b/components/navbar/Navbar.js
@@ -5,8 +5,10 @@ import NavLink from "./NavLink";
 import Link from "../Link";
 import app from "../../config/app.json";
 import Image from "next/legacy/image";
-import { FaGithub } from "react-icons/fa";
-import { IconContext } from "react-icons";
+import getIcon from "../Icon";
+import { IconContext } from "react-icons/lib/cjs/iconContext";
+
+const FaGithub = getIcon("FaGithub");
 
 export default function Navbar() {
   const [isOpen, setIsOpen] = useState(false);
--- a/components/navbar/Navbar.js
+++ b/components/navbar/Navbar.js
@@ -6,8 +6,6 @@ import Link from "../Link";
 import app from "../../config/app.json";
 import Image from "next/legacy/image";
 import getIcon from "../Icon";
-import { IconContext } from "react-icons/lib/cjs/iconContext";
-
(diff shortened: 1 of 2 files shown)


stevenovak123 (dismissed):

components/navbar/Navbar.js
```diff
@@ -5,8 +5,10 @@ import NavLink from "./NavLink";
 import Link from "../Link";
 import app from "../../config/app.json";
 import Image from "next/legacy/image";
-import { FaGithub } from "react-icons/fa";
-import { IconContext } from "react-icons";
+import getIcon from "../Icon";
+import { IconContext } from "react-icons/lib/cjs/iconContext";
```
eddiejaoude, line 9: Changing this import throws an error for using `value` for styles

Screenshot 2023-02-04 at 14 54 57

dan-mba: Ugh, a hydration error. I will have a look at it.

eddiejaoude: thank you 👍

dan-mba: I removed IconContext & added a tailwind class to fix the problem.
FYI, vertical-align is irrelevant for a svg element so I didn't add it with tailwind.

eddiejaoude: oh I thought the `IconContext` was needed to add styles to the icon (not sure why I thought that)

Thanks for fixing, I will have a quick play and merge in the morning

dan-mba: `IconContext` is only needed if something needs to be applied directly to the svg & can't be applied to a wrapper element.
In this case because the svg sets `stroke` & `fill` to `currentColor` then it will use the inherited color.

eddiejaoude: oh I see, thank you 👍

eddiejaoude (approved):
Looks good 👍 

I wonder if every import of `react-icons/*` should be converted?

dan-mba:
> I wonder if every import of `react-icons/*` should be converted?

Probably, but I picked the profile page & _app because the icons had the most effect on initial JavaScript load.

eddiejaoude:
Sure 👍 the others like `Alerts` etc can come in another PR

PR #4663, opened 2023-02-10
Fix daily stats init

dan-mba:
## Changes proposed
When creating the initial user record we don't save the result.
The _id field is used later, but for the initial user record this wouldn't exist.
This PR saves the returned document from the initial creation.
It also fixes some logging spelling mistakes.

### DIFF
--- a/pages/api/users/[username]/index.js
+++ b/pages/api/users/[username]/index.js
@@ -29,11 +29,11 @@ export default async function handler(req, res) {
   const date = new Date();
   date.setHours(1, 0, 0, 0);
 
-  const getProfile = await Profile.findOne({ username });
+  let getProfile = await Profile.findOne({ username });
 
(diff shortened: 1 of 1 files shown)


pages/api/users/[username]/index.js
```diff
@@ -29,11 +29,11 @@ export default async function handler(req, res) {
   const date = new Date();
   date.setHours(1, 0, 0, 0);
 
-  const getProfile = await Profile.findOne({ username });
+  let getProfile = await Profile.findOne({ username });
 
   if (!getProfile) {
     try {
-      await Profile.create({
+      getProfile = await Profile.create({
```
Comment (eddiejaoude) [inline pages/api/users/[username]/index.js:36]: Good idea 👍

eddiejaoude (approved):
Thanks 👍

PR #4671, opened 2023-02-10
Check provided location for remote

dan-mba:
## Changes proposed
Check for remote must be on provided

### DIFF
--- a/pages/map.js
+++ b/pages/map.js
@@ -19,7 +19,7 @@ export async function getServerSideProps() {
       user.location &&
       (user.location.provided !== "unknown" &&
         user.location.name !== "unknown" &&
-        user.location.name.toLowerCase() !== "remote")
+        user.location.provided.toLowerCase() !== "remote")
   );
 
   return {


pages/map.js
```diff
@@ -19,7 +19,7 @@ export async function getServerSideProps() {
       user.location &&
       (user.location.provided !== "unknown" &&
         user.location.name !== "unknown" &&
-        user.location.name.toLowerCase() !== "remote")
+        user.location.provided.toLowerCase() !== "remote")
```
amandamartin-dev, line 22: Do we need to check name and provided similar to the "unknown" case or will this not matter for remote?

amandamartin-dev: Also - should we add toLowerCase() to the unknown scenario as well or would that never be capitalized as it's likely auto-generated versus user provided?

dan-mba: The default return object is
```
  let location = {
    provided: "unknown",
    name: "unknown",
    lat: 0,
    lon: 0,
  };
```

This is returned if the GitHub lookup fails.
My first change also returns this if GitHub returns `remote`

If the API lookup fails, provided is changed to what GitHub returned & the rest is sent back unchanged.

If the API is sent `remote`  it return Remote ... Oregon which won't match the previous check

amandamartin-dev (approved):

PR #4684, opened 2023-02-11
feat: account cards

eddiejaoude:
Ignore numbers, my test data is a bit messy

Screenshot 2023-02-11 at 11 10 22

### DIFF
--- a/pages/account/statistics.js
+++ b/pages/account/statistics.js
@@ -16,6 +16,7 @@ import Alert from "../../components/Alert";
 import Page from "../../components/Page";
 import PageHead from "../../components/PageHead";
 import { abbreviateNumber } from "../../services/utils/abbreviateNumbers";
+import BasicCards from "../../components/statistics/BasicCards";
 
 export async function getServerSideProps(context) {
   const session = await unstable_getServerSession(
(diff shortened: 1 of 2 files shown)


pages/account/statistics.js
```diff
@@ -16,6 +16,7 @@ import Alert from "../../components/Alert";
 import Page from "../../components/Page";
 import PageHead from "../../components/PageHead";
 import { abbreviateNumber } from "../../services/utils/abbreviateNumbers";
+import BasicCards from "../../components/statistics/BasicCards";
 
 export async function getServerSideProps(context) {
   const session = await unstable_getServerSession(
```
krshkun, line 22: Just as a reminder, on newer versions of next-auth this function is not unstable, so we might want to update ✅

krshkun (commented):
I did not understand what exactly was changed, but I do not think I see something that breaks anything, waiting for other reviews 👍

pages/account/statistics.js
eddiejaoude: good spot, yes this has been deprecated 👍 we will need a separate PR to update all of these

eddiejaoude:
> I did not understand what exactly was changed, but I do not think I see something that breaks anything, waiting for other reviews 👍

Thanks for reviewing 👍  the changes are the 3 boxes at the top of the page above the graph

PR #4898, opened 2023-02-17
fix: Markdown not being rendered in the description when sharing profile

### LINKED ISSUE #3950: Fix social media preview of LinkFree profile
### Description

I shared my linkfree profile URL with a friend and I spotted - in profile preview the bio tag appears as markdown.
It would be good if it appears as simple text without the markdown symbols.

### Screenshots

WhatsApp Image 2023-01-25 at 13 40 53
Have a look below my name - asterisk symbols also appear.

### Additional information

I don't know is it a bug or whatever! I thought it could be improved - so I learned how to refer an issue in an open source project!

ChinmayMhatre:
## Fixes Issue
closes #3950 

## Changes proposed

Remove markdown from the description for now. as discussed in the issue. maybe markdown could be added if any new solution is found.

<a href="https://gitpod.io/#https://github.com/EddieHubCommunity/LinkFree/pull/4898"></a>

### DIFF
--- a/pages/[username].js
+++ b/pages/[username].js
@@ -88,7 +88,7 @@ export default function User({ data, BASE_URL }) {
     <>
       <PageHead
         title={data.name}
-        description={data.bio}
+        description={data.bio.replaceAll("*","")}
         ogTitle={data.name}
         ogUrl={`https://linkfree.eddiehub.io/${data.username}`}
         ogImage={data.avatar}
--- a/pages/[username].js
+++ b/pages/[username].js
@@ -17,6 +17,15 @@ import UserTestimonials from "../components/user/UserTestimonials";
 import UserEvents from "../components/user/UserEvents";
 import Page from "../components/Page";
 
+import { remark } from 'remark'
+import strip from 'strip-markdown'
+
+const convertToText = async (markdown) => {
+  const textOutput = await remark()
+    .use(strip)
+    .process(markdown)
+  return String(textOutput)
+}
 export async function getServerSideProps(context) {
   const { req } = context;
   const username = context.query.username;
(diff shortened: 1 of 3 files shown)


sansarmaske (dismissed):
Works as expected.

pages/[username].js
```diff
@@ -88,7 +88,7 @@ export default function User({ data, BASE_URL }) {
     <>
       <PageHead
         title={data.name}
-        description={data.bio}
+        description={data.bio.replaceAll("*","")}
```
Comment (eddiejaoude) [inline pages/[username].js:91]: we will also need to remove other types of markdown that people might use

eddiejaoude (changes requested):
Looks good, but left an inline comment

pages/[username].js
Comment (ChinmayMhatre) [reply pages/[username].js:91]: Oh yeah! I missed that. I'll try to create a function before passing in to `description`

Comment (eddiejaoude) [reply pages/[username].js:91]: It might be easier to do a regex that extracts the text and punctuation

ChinmayMhatre:
Would using something like this plugin be okay?

https://www.npmjs.com/package/strip-markdown

pages/[username].js
Comment (ChinmayMhatre) [reply pages/[username].js:91]: Oh, like just letters, spaces and commas?

Comment (ChinmayMhatre) [reply pages/[username].js:91]: I looked around for this. We can certainly do this with regex. It may not take into account all the complex cases. I used chatgpt to create it for as many cases as we can. We can try this for now and add an external library if this fails

ChinmayMhatre:
Could we use something like this : https://www.npmjs.com/package/strip-markdown

I think regex wouldn't take into account complex cases. What do you think?

eddiejaoude:
yeah that could work 👍

eddiejaoude (approved):
Looks good, thank you 👍

eddiejaoude:
I think there is a promise issue...

Screenshot 2023-02-23 at 08 19 48

eddiejaoude:
I have fixed in this PR https://github.com/EddieHubCommunity/LinkFree/pull/5022

PR #4997, opened 2023-02-21
Call username lookup code directly

dan-mba:
## Changes proposed
Our current method of getting data in `getServerSideProps` is an [anti-pattern](https://nextjs.org/docs/basic-features/data-fetching/get-server-side-props#getserversideprops-or-api-routes)

This PR calls the functionality from `/api/users/[username]` API directly instead of using a fetch.

<a href="https://gitpod.io/#https://github.com/EddieHubCommunity/LinkFree/pull/4997"></a>


pages/[username].js
```diff
@@ -24,14 +20,12 @@ export async function getServerSideProps(context) {
...
+  if(status === 200) {
+    data = userData;
     log.info(`data loaded for username: ${username}`);
-  } catch (e) {
-    log.error(e, `profile loading failed for username: ${username}`);
+  } else {
+    log.error(userData.error, `profile loading failed for username: ${username}`);
   }
 
   if (!data.username) {
```
Comment (eddiejaoude) [inline pages/[username].js:31]: I wonder if this could be moved to also remove the `else`

```js
if (!data.username || status !== 200) {
```

pages/api/users/[username]/index.js
```diff
@@ -149,7 +160,14 @@ export default async function handler(req, res) {
...
+  const retData = {
+    status: 200,
+    userData: {
+      username,
+      ...data,
+      location: profileWithLocation.location
+    }
+  };
+  
+  return JSON.parse(JSON.stringify(retData));
```
Comment (eddiejaoude) [inline pages/api/users/[username]/index.js:172]: Why do we need to do both?

pages/[username].js
Comment (dan-mba) [reply pages/[username].js]: If  username is not found, status will not be 200.
So I combined the fail paths into one.

pages/api/users/[username]/index.js
Comment (dan-mba) [reply pages/api/users/[username]/index.js:171]: The reason for this is to return an object that can be translated to JSON.
If I don't do this, Next complains when I try to include this object as part of the data returned from `getServerSideProps`.
I think it has something to do with converting Date structures to something JSON can handle.

Comment (eddiejaoude) [reply pages/api/users/[username]/index.js:171]: oh I see, thanks for explaining

eddiejaoude:
Looks good Dan 👍  I am keen to test it out

note: If we can merge when I am at my computer just incase anything needs to be rollback on the server etc

eddiejaoude:
We should make an issue (epic label) and list other areas that need updating to be consistent

dan-mba:
I created an issue that lists the pages that need to be updated

eddiejaoude:
Thanks @dan-mba https://github.com/EddieHubCommunity/LinkFree/issues/5069

eddiejaoude (approved):
Looks good 👍

Issue #5069, opened 2023-02-25
Replace fetches from internal APIs with function calls


dan-mba:
### Description

The [Next.js documentation](https://nextjs.org/docs/basic-features/data-fetching/get-server-side-props#getserversideprops-or-api-routes) explains that fetching data from an API route from the same server is a bad idea.

This was fixed for user profiles in #4997 .

The following pages need to be updated to follow the same style:
- [x] pages/map.js - #6939 
- [x] pages/events.js - #5152 
- [x] pages/account/statistics.js - #7028 
- [x] pages/index.js - #5124 
- [x] pages/search.js - #6946

Harrish-Selvarajah:
Hi @eddiejaoude and @dan-mba. I see that map.js, statistic.js and search.js files need to be replaced with function calls. I'm open to take this up :).

dan-mba:
> Hi @eddiejaoude and @dan-mba. I see that map.js, statistic.js and search.js files need to be replaced with function calls. I'm open to take this up :).

@Harrish-Selvarajah Sorry I missed your comment.
Work got stalled on this issue by the migration of loading profiles from the database.
Then got combined with fixing some performance issues.
Since there is only one file left, I will handle it to get this issue closed.
Again, I apologize for not responding sooner.

Harrish-Selvarajah:
Hey @dan-mba, No worries! 😀

dan-mba:
All pages are migrated

PR #5152, opened 2023-02-28
Convert fetches to function calls for events page

dan-mba:
## Fixes Issue
Fixes events page for #5069 

<a href="https://gitpod.io/#https://github.com/EddieHubCommunity/LinkFree/pull/5152"></a>


pages/api/events.js
```diff
@@ -53,5 +67,10 @@ export default async function handler(req, res) {
     .filter((event) => new Date(event.date.end) > new Date())
     .sort((a, b) => new Date(a.date.start) - new Date(b.date.start));
 
-  res.status(200).json(eventsFiltered);
+  return {
+    statusCode: 200,
+    events: {
+      eventsArray: eventsFiltered
+    }
```
eddiejaoude, line 74: Having this nested with the type I don't think is consistent with similar parts of the code?

dan-mba: The problem is this is the only one I have updated so far that didn't return an object.

In order to have a place to put an error message, I return an object.
The object contains the array on success or the error message on failure

eddiejaoude: ah ok, what about having the error message outside the data, similar to `status`?

dan-mba: Doing it that way would make it more difficult to work with the existing API. 

It make more sense for me for an API to return a status code & JSON payload.
This puts the responsibility on the caller to check the status code & evaluate the JSON payload depending on that status.

eddiejaoude (approved):
Looks good, thank you 👍

PR #5534, opened 2023-03-18
fix: remove duplicate tags in search when case is different

### LINKED ISSUE #4676: Tag duplication when case is different on search page
### Description

On the search page there are 3 different tags for JS due to different use of case.  Recommend standardizing all tags (not just JS) into the same case to prevent duplication

search bug
 

### Screenshots

_No response_

### Additional information

_No response_

joe-shajan:
## Fixes Issue

 Closes #4676

To fix this issue I  converted all the tags to lowercase and passed it to the component.

### Before
search before

### After
after

## Note to reviewers

<a href="https://gitpod.io/#https://github.com/EddieHubCommunity/LinkFree/pull/5534"></a>


joeshajan:
Hey @eddiejaoude what do you think about this solution?

joeshajan:
Hey @amandamartin-dev can you take a look at it

eddiejaoude:
Sorry for my slow reply, it is on my list to do soon

eddiejaoude:
> Hey Eddie I checked this and the search looks good, but I noticed on the map that it appears not to have the same code currently on production so I"m not sure if something is up or it's just an oddity of gitpod. The things i noticed were the map bounds and the clear filters code seems to be missing from this

hi @amandamartin-dev thank you for checking. It will be because the branch is not up to date. I will do an update

pages/search.js
```diff
@@ -122,8 +156,8 @@ export default function Search({ data }) {
         <h1 className="text-4xl mb-4 font-bold">Search</h1>
 
         <div className="flex flex-wrap justify-center mb-4">
```
CBID2, line 158:
```suggestion
        <article className="flex flex-wrap justify-center mb-4">
```

CBID2 (changes requested):
Hi @joeshajan! 😄 I made a suggestion on how to improve this line of code. Consider using `<div>` tags sporadically because they can make the code inaccessible.

joeshajan:
> Hi @joeshajan! 😄 I made a suggestion on how to improve this line of code. Consider using `<div>` tags sporadically because they can make the code inaccessible.

Thanks 😊

CBID2 (dismissed):
LGTM!

amandamartin-dev (changes requested):
Hello! You have a build failure in this file here due to changing only the opening tag to a different type:

"./pages/search.js
181:8  Error: Parsing error: Expected corresponding JSX closing tag for < article >. (181:8)"

If you need assistance on how to fix this please let us know!

joe-shajan:
> Hello! You have a build failure in this file here due to changing only the opening tag to a different type:
> 
> "./pages/search.js 181:8 Error: Parsing error: Expected corresponding JSX closing tag for < article >. (181:8)"
> 
> If you need assistance on how to fix this please let us know!

Sorry, I didn't notice that. I have made the needed changes.

pages/search.js
```diff
@@ -128,9 +162,9 @@ export default function Search({ data }) {
       <Page>
         <h1 className="text-4xl mb-4 font-bold">Search</h1>
 
-        <div className="flex flex-wrap justify-center space-x-3 mb-4">
-          {tags &&
-            tags
+        <article className="flex flex-wrap justify-center mb-4">
```
amandamartin-dev, line 165: Sorry to be a bit picky here, but I'm trying to understand the change to an article html component from a div.  @EmmaDawsonDev can you weigh in on if this makes sense semantically? I'm thinking no becuase it's a group of tags. I would think leaving div would have been appropriate here but not sure for something like this.

pages/search.js
```diff
@@ -26,7 +26,41 @@ export async function getServerSideProps(context) {
     const res = await fetch(
       `${process.env.NEXT_PUBLIC_BASE_URL}/api/discover/tags`
     );
-    data.tags = await res.json();
+
+    const tags = await res.json();
+
+    /*
```
amandamartin-dev, line 32: I think the code is self explanatory in this case and perhaps we do not need the comment?  Appreciate this though! @eddiejaoude wdyt? keep the comment, or would you understand what is going on here without it?

pages/search.js
joe-shajan: https://github.com/EddieHubCommunity/LinkFree/pull/5534#pullrequestreview-1368018671

EmmaDawsonDev: > Sorry to be a bit picky here, but I'm trying to understand the change to an article html component from a div.  @EmmaDawsonDev can you weigh in on if this makes sense semantically? I'm thinking no becuase it's a group of tags. I would think leaving div would have been appropriate here but not sure for something like this.

Sorry for the late reply! Article should be used for something self contained when something can be completely lifted out from where it is, moved to another place and still make sense like a blog article. I don't think it makes sense in this case and a div is better.

eddiejaoude: > #5534 (review)

@joe-shajan I agree this should remain as a `div` 👍  plus it will be consistent with the rest of the app (btw don't make every change that anyone suggests on your PR, if you disagree or unsure question the reason for the comment, or wait for the maintainers to confirm if the change is required)

pages/search.js
eddiejaoude: Yes I agree, I think the comments can be removed

joe-shajan: 👍done

pages/search.js
```diff
@@ -26,7 +26,27 @@ export async function getServerSideProps(context) {
...
+          ...tagsWithOutDuplicates[lowerName],
+          total: tagsWithOutDuplicates[lowerName].total + total,
+        };
+      } else {
+        tagsWithOutDuplicates[lowerName] = {
+          name: lowerName,
+          total,
+        };
+      }
+    });
```
eddiejaoude, line 47: I think this could be done with a `reduce` that would not require the `if/else`

(not tested)
```js
const uniqueTags = tags.reduce((allTags, tag) => {
  const lowerTag = tag.toLowerCase();
  const currCount = allTags[lowerTag] ?? 0;
  return {
    ... allTags,
    [tag]: currCount + 1,
  };
}, {});
```

eddiejaoude: Also, now we are getting the data from the database, we could use mongo's `group` and `count` feature in the `aggregation`

- https://www.mongodb.com/docs/manual/core/aggregation-pipeline/?_ga=2.243062658.1416307675.1684809031-436223288.1684493318
- https://www.mongodb.com/docs/manual/reference/operator/aggregation/group/?_ga=2.245751556.1416307675.1684809031-436223288.1684493318
- https://www.mongodb.com/docs/manual/reference/operator/aggregation/count/?_ga=2.245751556.1416307675.1684809031-436223288.1684493318

shobhitexe: Hey i dont know if i am supposed to write here or not 
sorry if i am not 

Have find a quick fix using aggregator as eddie mentioned 

i see aggregate is already implemented on [tags.js](https://github.com/EddieHubCommunity/LinkFree/blob/main/pages/api/discover/tags.js) file 

just if we convert all tags to lowercase it will filter all 

image

The fix i have found lemme know if this works or do i have to go more deep into this 
Thanks

SaraJaoude: @shobhitexe thank you for the suggestion.  This does look like a possible solution (we have not tested this) but in addition when the search is being performed using the selected tag it would also need to be converted into lower case otherwise it would not return any results.

SaraJaoude:
Due to lack of activity I will be closing this Pull Request.

pages/search.js
shobhitexe: Yes correct got it

PR #5765, opened 2023-04-01
feat: load json files into the DB and app uses DB

### LINKED ISSUE #5575: [FEATURE] Manage profile with a form
### Description

Page to manage profile of authenticated user

- toggle switch to use json file or form
- profile (name, bio, etc)
- links
- social shortcuts
- events
- testimonials

Conditions
- profile form with json file data if new (not existing)
- display alert if json file and MongoDB entry both exist - recommend they remove the json file

### TODO

- [x] manage page
- [x] update basic profile
- [x] form tags
- [x] ~~toggle between file and database source~~ no longer needed
- [x] add manage link to navbar
- [ ] ~~toggle profile enable/disable~~ low priority
- [ ] ~~associate profile to the logged in account~~ moved to another issue https://github.com/EddieHubCommunity/LinkFree/issues/7065
- [ ] ~~if username is different from logged in user, show a warning and have a button to update (this will only work if someone logged in before they changed their username)~~ moved to separate issue https://github.com/EddieHubCommunity/LinkFree/issues/7065
- [x] warning if json file exists
- [x] form links + socials
- [x] form testimonials
- [x] form milestones
- [x] form events

**notes:**
- must be able to handle multiple profiles in the database
- rename the profile (username) if available

### Screenshots

_No response_

### Additional information

_No response_

### LINKED ISSUE #6244: [FEATURE] Load all json profiles into Mongo on project start
**NOTE: profiles that have `social links` that are not in their `links` section will not appear**

### Description

We are adding forms for people to manage their profiles if they prefer over json files, but having the application switch between both is more code to manage and test etc

An idea is to load all json profiles into the database, there for the app will always display profiles from the database, even if they are managed by a json file or by the form

**Part 1**
- [x] load all json files
   - [x] load all testimonials even the ones not included on the profile
   - [x] but then set `enabled` flag for testimonials included on profile
   - [x] flag links that have social shortcuts (throw error if doesn't exist)
- [x] upsert into DB on app start
   - [x] links + socials
   - [x] milestones
   - [x] testimonials + pinned
   - [x] events
- [x] app reads all profiles from DB
   - [x] Profile
      - [x] links + socials
      - [x] milestones
      - [x] testimonials + pinned
      - [x] events
   - [x] tags
   - [x] Search ~~(+ query parameter?)~~ do this later 
   - [x] Map
   - [x] Events

**Part 2**
- [x] Secret for GitHub Action + Vercel
- [x] API reload checks environment variable equals incoming secret
- [x] GitHub Action: Hit API end point after deployment to load in json files (secure with secret)
- [x] Deploy to Preview environment

```yml
  load-json-profies:
    if: github.event_name == 'deployment_status' && github.event.deployment_status.state == 'success'
    run: curl https://linkfree.io/api/system/reload?token=XXX-YYY-ZZZ
 ```

Reference https://vercel.com/guides/how-can-i-run-end-to-end-tests-after-my-vercel-preview-deployment

**Part 3**

Separate PR https://github.com/EddieHubCommunity/LinkFree/pull/6875

- [ ] profiles managed by forms set a `form` flag, when loading all json files skip profiles that have this flag set
- [ ] on user's dashboard if viewing stats or editing their profile that has form flag set and there is a json file, show warning to remove json file as it is no longer used

### Screenshots

_No response_

### Additional information

_No response_

### PR DESCRIPTION
**Pull request by eddiejaoude**
## Fixes Issue

closes #6244 

- [x] ~~note: please ignore the new form files, I need to move these to another PR~~ removed

- Q. Deploy to preview environment on Vercel to test?
   - Data going into `prod` DB should not affect the current app?
   - Or should we use a `preview` DB (easily setup on Atlas)

<a href="https://gitpod.io/#https://github.com/EddieHubCommunity/LinkFree/pull/5765"></a>

### CODE REVIEW

**Comment by eddiejaoude** [comment]
I can't replicate the failing test locally, I will look into it tomorrow

File: pages/api/system/reload.js [hunk 1]
```diff
@@ -0,0 +1,279 @@
...
+
+  // 1. load all profiles
+  const basicProfiles = findAllBasic();
+  const fullProfiles = basicProfiles.map((profile) =>
+    findOneByUsernameFull(profile)
+  );
+
+  // 2. save basic profiles to database
+  // only if `source` is not `database` (this will be set when using forms)
+  fullProfiles.map(async (profile) => {
```
dan-mba, line 29: This section need to be inside a `Promise.all()` which you will need to await. Otherwise the function will return before the code is finished running.

eddiejaoude: I thought the promise all is only required if I didn't `await` for each one and wanted them done in parallel? My understanding is at the moment it does one at a time and waits for each one t complete before moving to the next one

eddiejaoude: For example (simplified)...

**Current sequentially**
```js
fullProfiles.map(async (profile) => 
   await Profile.findOne({
      username: profile.username,
   });
);
```

VS

**In parallel**

```js
await Promise.all(fullProfiles.forEach((profile) => 
   Profile.findOne({
      username: profile.username,
   });
));
```

dan-mba: First off, all async functions return a promise.

Using `array.map()` or `array.forEach()` with an async callback function means they can run concurrently (they won't truly run in parallel as node is single threaded by default).

From my testing, if I limit the fullProfiles array to 100 users, everything works fine.
But when I use the full array, I get the same errors the tests get. 
I believe this is caused by database requests backing up and all the promises are returned before the functions are complete.

Thus, using `Promise.all()` will ensure everything is complete before returning.

eddiejaoude: > all async functions return a promise.

oh that is a good point 👍 


Thanks for testing, I will update 👍

.github/workflows/vercel-beta.yml
```diff
@@ -0,0 +1,27 @@
...
+        run: vercel pull --yes --environment=preview --token=${{ secrets.VERCEL_TOKEN }}
+      - name: build project artifacts
+        run: vercel build --token=${{ secrets.VERCEL_TOKEN }}
+      - name: deploy project artifacts to Vercel
+        run: vercel deploy --prebuilt --token=${{ secrets.VERCEL_TOKEN }}
+      - name: assign beta domain
+        run: |
+          vercel deploy --prebuilt --token=${{ secrets.VERCEL_TOKEN }} > domain.txt
+      #    vercel alias --token=${{ secrets.VERCEL_TOKEN }} set `cat domain.txt` beta-linkfree.vercel.app
+      - name: load json files
```
dan-mba, line 26: Is there a reason to do this on all deploys?
Shouldn't this be behind a check to see if a file in the data dir was modified?

.github/workflows/vercel-beta.yml
```diff
@@ -0,0 +1,27 @@
...
+        run: curl -f https://linkfree-eddiejaoude-eddiehub.vercel.app/api/system/reload?secret=${{ secrets.LINKFREE_API_SECRET }}
```
dan-mba, line 27: Is there a way to load `LINKFREE_API_SECRET` using the vercel command?
This would make it easier than keeping the same value in 2 places.

.github/workflows/vercel-beta.yml
eddiejaoude: Oh good point, that would be more efficient 💯

.github/workflows/vercel-beta.yml
eddiejaoude: Yes true, this could come later as an improvement I think

pages/api/system/reload.js
```diff
@@ -50,9 +50,40 @@ export default async function handler(req, res) {
           bio: profile.bio,
           tags: profile.tags,
         },
-        { upsert: true }
+        { upsert: true, new: true }
       );
 
+      // add social urls to links but disable them
+      try {
+        if (profile.links && profile.socials) {
```
eddiejaoude, line 58: `profile.links` may not exist

.github/workflows/vercel-beta.yml
```diff
@@ -0,0 +1,41 @@
+name: Deploy beta branch to Vercel Preview
+env:
+  VERCEL_ORG_ID: ${{ secrets.VERCEL_ORG_ID }}
+  VERCEL_PROJECT_ID: ${{ secrets.VERCEL_PROJECT_ID }}
+on:
+  workflow_run:
+    workflows: [Build]
```
eddiejaoude, line 7: Did not trigger?

eddiejaoude: On release?

PR #4850, opened 2023-02-15
Feat: Search Bar Story added

ThePiyushAggarwal:
## Fixes Issue
Closes #1989 

## Changes proposed

- [x] Refactored Search Component into a separate file for reuse in story
- [x] Changed the placeholder in Search Bar for better grammar
- [x] Added the Search Bar Story

## Check List (Check all the applicable boxes) 

- [x] My code follows the code style of this project.
- [ ] My change requires changes to the documentation.
- [ ] I have updated the documentation accordingly.
- [ ] All new and existing tests passed.
- [x] This PR does not contain plagiarized content.
- [x] The title of my pull request is a short description of the requested changes.

## Screenshots

image

<a href="https://gitpod.io/#https://github.com/EddieHubCommunity/LinkFree/pull/4850"></a>

PR #5785, opened 2023-04-02
Feat: Search Bar Story added

### LINKED ISSUE #1989: [FEATURE] Add page pages/search.js to Storybook
### Description
All React components and pages need to be added to Storybook (in the nextjs branch)

> Note
> Each item listed below for component/page should be in a new Issue and PR

- [ ] pages/search.js

### Screenshots

_No response_

### Additional information

#1944 search component in the nextjs branch for storybook.

ThePiyushAggarwal:
## Fixes Issue
Closes #1989 
PR #4850 is an old PR which contains some explaination for changes made in this PR.

## Changes proposed

- [x] Refactored Search Component into a separate file for reuse in story
- [x] Changed the placeholder in Search Bar for better grammar
- [x] Added the Search Bar Story
- [x] Added autofocus on search input, which was originally there. It was removed while [creating a common Input component](https://github.com/EddieHubCommunity/LinkFree/pull/5273/files#diff-99c55e6229f7a0d904e59da5da6e77729e62b05852484cf2afff2e13292b50b1), probably because of some error that the developer had to face.

## Screenshots

image

<a href="https://gitpod.io/#https://github.com/EddieHubCommunity/LinkFree/pull/4850"></a>

<a href="https://gitpod.io/#https://github.com/EddieHubCommunity/LinkFree/pull/5785"></a>

### DIFF
--- a/components/SearchInput.js
+++ b/components/SearchInput.js
@@ -1,0 +1,22 @@
+import { useEffect, useRef } from "react";
+
+import Input from "./form/input";
+
+export default function SearchInput({ value, onChange }) {
+  const inputRef = useRef();
+
+  // Makes autofocus work on search input
+  useEffect(() => {
+    inputRef.current.focus();
+  }, []);
+
+  return (
+    <Input
+      placeholder="Search user by name or tags, e.g., open source, reactjs"
+      inputRef={inputRef}
+      name="keyword"
+      value={value}
+      onChange={(e) => onChange(e.target.value)}
+    />
+  );
+}
--- a/stories/components/SearchInput.stories.js
+++ b/stories/components/SearchInput.stories.js
@@ -1,0 +1,14 @@
+import { useState } from "react";
+import SearchInput from "../../components/SearchInput";
+
+export default {
+  component: SearchInput,
+  title: "components/Search Input",
+};
+
+const Template = (args) => {
+  const [value, setValue] = useState("");
+  return <SearchInput {...args} value={value} onChange={setValue} />;
+};
+
+export const Basic = Template.bind({});
--- a/components/form/input.js
+++ b/components/form/input.js
@@ -1,4 +1,10 @@
-export default function Input({ name, value, placeholder, ...restProps }) {
+export default function Input({
+  name,
+  inputRef,
+  value,
+  placeholder,
+  ...restProps
+}) {
   return (
     <input
       type="text"
@@ -6,6 +12,7 @@ export default function Input({ name, value, placeholder, ...restProps }) {
       className="border-2 hover:border-orange-600 transition-all duration-250 ease-linear rounded px-6 py-2 mb-4 block w-full"
       name={name}
       value={value}
+      ref={inputRef}
       {...restProps}
     />
   );
--- a/pages/search.js
+++ b/pages/search.js
@@ -1,4 +1,4 @@
-import { useEffect, useState, useRef } from "react";
+import { useEffect, useState } from "react";
 import { useRouter } from "next/router";
 
 import UserCard from "@components/user/UserCard";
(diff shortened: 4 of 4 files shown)


components/SearchInput.js
```diff
@@ -0,0 +1,22 @@
...
+  return (
+    <Input
+      placeholder="Search user by name or tags, e.g., open source, reactjs"
+      inputRef={inputRef}
+      name="keyword"
+      value={value}
+      onChange={(e) => onChange(e.target.value)}
+    />
+  );
+}
```
eddiejaoude, line 22: I don't think this file is needed, seems like too much abstraction to me as we are unlikely to reuse this?

stories/components/SearchInput.stories.js
```diff
@@ -0,0 +1,14 @@
...
+  component: SearchInput,
+  title: "components/Search Input",
+};
+
+const Template = (args) => {
+  const [value, setValue] = useState("");
+  return <SearchInput {...args} value={value} onChange={setValue} />;
+};
+
+export const Basic = Template.bind({});
```
eddiejaoude, line 14: Looks good, but I think the Story for the `input` would be best

components/SearchInput.js
ThePiyushAggarwal: Earlier this was required since there was no Input component. Now, I think there is only one difference which is autofocus on the Input box. Do you suggest I should put the autofocus as a boolean prop on the `Input` component and use?

stories/components/SearchInput.stories.js
ThePiyushAggarwal: So, the issue #1989 isn't valid anymore. Should I open another issue for `Input` story and make this PR work that way?

components/SearchInput.js
eddiejaoude: But the `Input` component has `restProps` so it should be ok I think?

stories/components/SearchInput.stories.js
eddiejaoude: Yes #1989 is still required but I don't understand why this PR has a new component, it is better to have PR's focused and I still think this PR is trying to make too many changes (I really believe less is more)

components/SearchInput.js
ThePiyushAggarwal: No, the `Input` component will have its own `ref` and I will do the autofocus based on the boolean prop inside the `useEffect`. That's how it will work if we want to use a prop

stories/components/SearchInput.stories.js
ThePiyushAggarwal: The issue was to create a story for the `Search` component. So, the only way to use the same `Search` component was to have it in such a way that it can be imported in the story.
Now, when the new `Input` component is also there, I also think that the `Search` component isn't doing much. So, I am suggesting to have a `autoFocus` prop on the `Input` component.
I am suggesting I can create a new issue because we can just create a `Input` story and have a variation under that with a `Search Input` label which uses a different placeholder and the `autoFocus` is also turned on.
I can put both stories in the same PR because it will add only one line and it will solve two issues.

eddiejaoude: My understanding of the issue was to add a Story for the Search page

ThePiyushAggarwal: Sorry, I didn't read the issue properly. I will create another PR for this issue.

eddiejaoude: Ok no problem, sorry for the confusion. Usually we prefer the minimum amount of changes and any further improvements discovered by doing the work can come in another issue and PR. Thank you for understanding 👍

PR #5797, opened 2023-04-02
added tailwind classes to home page

ChinmayMhatre:
image

image

image

### DIFF
--- a/components/Testimonials.js
+++ b/components/Testimonials.js
@@ -2,10 +2,10 @@ import Link from "./Link";
 
 export default function Testimonials({ data }) {
   return (
-    <div className="bg-white dark:bg-gray-800 pt-0 pb-24">
+    <div className="bg-white dark:bg-primary-high pt-0 pb-24">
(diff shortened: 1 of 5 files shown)


components/Testimonials.js
```diff
@@ -28,7 +28,7 @@ export default function Testimonials({ data }) {
                     />
                     <div>
                       <Link
-                        className="font-semibold text-gray-900 dark:text-gray-300"
+                        className="font-semibold text-primary-high dark:text-gray-300"
```
eddiejaoude, line 31: I think this `dark:text-gray-300` should be one of our pre-made classes?

eddiejaoude:
This looks great 👍  I left just 1 small comment

components/Testimonials.js
eddiejaoude: I will merge and this can be fixed in the next PR

eddiejaoude: Also I think the buttons on the homepage still have numbers or their colours and not the new classes

PR #5841, opened 2023-04-04
Closes #5812 append hover:underline to the final result in components…

### LINKED ISSUE #5812: Add a visual effect on hover
### Description

On the events page, there are certain events mentioned. But the whole event box is not clickable and only the title of the box is such as `API Days Patna` in the screenshot. So, can we add a simple underline below the title. So, it will enhance the UX.
Screenshot (481)

### Screenshots

_No response_

### Additional information

_No response_

RutamBhagat:
## Fixes Issue

Adds a visual effect on hover

Closes #5812 

## Changes proposed

1. Add an underline on the Event name when mouse hover

## Screenshots

Before changes 
image

After changes
image

Note to reviewers
It's just a simple change that adds an underline to the event name when mouse is hovered over it

<a href="https://gitpod.io/#https://github.com/EddieHubCommunity/LinkFree/pull/5841"></a>

### DIFF
--- a/components/Link.js
+++ b/components/Link.js
@@ -5,9 +5,9 @@ export default function Link({ children, className, rel, ...restProps }) {
     <NextLink
       rel={rel ? rel : "noreferrer"}
       className={
-        className
+        `${className
           ? className
-          : "text-blue-600 underline decoration-dotted dark:text-blue-500 hover:underline hover:decoration-solid"
+          : "text-blue-600 underline decoration-dotted dark:text-blue-500 hover:underline hover:decoration-solid"} hover:underline`
       }
       {...restProps}
     >
--- a/components/Link.js
+++ b/components/Link.js
@@ -7,7 +7,7 @@ export default function Link({ children, className, rel, ...restProps }) {
       className={
         `${className
           ? className
-          : "text-blue-600 underline decoration-dotted dark:text-blue-500 hover:underline hover:decoration-solid"} hover:underline`
+          : "text-blue-600 underline decoration-dotted dark:text-blue-500 hover:decoration-solid"} hover:underline`
       }
       {...restProps}
     >


eddiejaoude:
Thank you for fixing 👍  Could you double check it works across the while app and there are no surprises? Thank you

RutamBhagat:
> Thank you for fixing +1 Could you double check it works across the while app and there are no surprises? Thank you

It works across the app but I think my previous solution of adding the hover:underline class only to the components/events/EventCard looked better in my opinion.

The current changes causes other components with Link tag have an underline as well such as the top nav bar buttons and the /docs page 
Screenshot from 2023-04-05 15-02-56
Screenshot from 2023-04-05 15-03-08

eddiejaoude:
I will have a look also, but I think being consistent across the app is probably better practice

amandamartin-dev:
I think to make this consistent with other changes to the same page, we should approach this in the same way that we are fixing #5870 which is to make the entire card clickable. 

@eddiejaoude and @RutamBhagat please let me know if you agree with that or would prefer to merge this in its' current state (added underline to clickable title on hover)

amandamartin-dev (commented):
This code creates some inconsistency on hover throughout other pages in the application.  If we decide to move forward with this merge, we should open subsequent issues to address the inconsistent underlined hover state that is now appearing on the docs page.

Also please see my other comment as there is a similar issue open for this page on mobile and collaboration on approach/consistency should be considered to make sure htis is the best way to address

eddiejaoude:
Sorry for my slow reply. Thanks Amanda for checking, I don't think it looks right across the app 🤔

eddiejaoude:
@RutamBhagat thank you for your contribution, but I am going to close this pull request and related issue, as I think the solution from another issue will be better for the app https://github.com/EddieHubCommunity/LinkFree/issues/5870

PR #6049, opened 2023-04-13
switch index page to incremental static regeneration

dan-mba:
## Changes proposed
Use incremental static regeneration to only build the index page every 15 minutes
This should lead to less CPU usage for the index page

<a href="https://gitpod.io/#https://github.com/EddieHubCommunity/LinkFree/pull/6049"></a>


Pradumnasaraf (commented):
Hey, @dan-mba

Can you explain why we are adding the Mongo service in the build step? Maybe I am missing something because we are just building the app and checking if anything is breaking or not.

dan-mba:
> adding

@Pradumnasaraf getStaticProps calls functions that access mongodb. Since this is part of the static build, mongodb must be available at build time, because the props are necessary for the build.

Pradumnasaraf:
Got it. Thanks for explaining. I have limited knowledge of the front end.

So, I think we also need to change the [`deploy.yml`](https://github.com/EddieHubCommunity/LinkFree/blob/main/.github/workflows/deploy.yml) as well.

Pradumnasaraf:
Got it. Thanks for explaining, @dan-mba . I have limited knowledge of the front end.

So, I think we also need to change the [deploy.yml](https://github.com/EddieHubCommunity/LinkFree/blob/main/.github/workflows/deploy.yml) as well.

dan-mba:
Now I have 2 questions:

Does the current Digital Ocean deployment use the container built by the deploy action?

How can we access the production MongoDB instance to build the container in the deploy action?

eddiejaoude:
> Now I have 2 questions:
> 
> Does the current Digital Ocean deployment use the container built by the deploy action?
> 
> How can we access the production MongoDB instance to build the container in the deploy action?

I don't think the docker container is used on DigitalOcean Apps, it does its own magic. But when we move back to Kubernetes we will use the Docker container

Pradumnasaraf:
Yes, LinkFree Digital Ocean deployment does look for the latest commit changes/hash, with no role of the GHCR image.

However, I wanted to clarify my previous question: will the `build.yml` action fail to start? This is because (@dan-mba) mentioned that our getStaticProps function needs MongoDB access. Yes, we are not currently the image.

dan-mba:
> However, I wanted to clarify my previous question: will the `build.yml` action fail to start? This is because you (@dan-mba) mentioned that our `getStaticProps` function needs MongoDB access.
> 

@Pradumnasaraf I added MongoDB to the build step in `build.yml` to allow the build process to complete. The database doesn't need to have data in it for the process to complete, it just need to be able to connect.

eddiejaoude:
> I added MongoDB to the build step in `build.yml` to allow the build process to complete. The database doesn't need to have data in it for the process to complete, it just need to be able to connect.

oh I didn't realise this, as I was going to add it won't have the data, but if that is not an issue 👍  - I need to learn more about this

Pradumnasaraf:
My bad @dan-mba . Instead of **deploy.yml** I said **build.yml**, sorry for that 🤦 

Again same question, because we use the build step in **deploy.yml** too.

>However, I wanted to clarify my previous question: will the ~`build.yml`~  `deploy.yml` action fail to start? This is because you (@dan-mba) mentioned that our `getStaticProps` function needs MongoDB access.

dan-mba:
> My bad @dan-mba . Instead of **deploy.yml** I said **build.yml**, sorry for that 🤦
> 
> Again same question, because we use the build step in **deploy.yml** too.
> 
> > However, I wanted to clarify my previous question: will the ~`build.yml`~  `deploy.yml` action fail to start? This is because you (@dan-mba) mentioned that our `getStaticProps` function needs MongoDB access.

Yes, we will need to add MongoDB in the `deploy.yml`, preferably using a secret for the prod database @Pradumnasaraf 
That is as long as we want to run `next build` in container build step.

For the index page, having slightly out of data stats is not a big problem.
But if we migrate this to other pages to improve performance, that might not be the case.

What would the penalty be of  creating a script for the CMD section of the Dockerfile that runs `next build` then `next start`?

eddiejaoude:
> For the index page, having slightly out of data stats is not a big problem.
But if we migrate this to other pages to improve performance, that might not be the case.

I don't think we have any pages that need near realtime data, so if other pages are 15mins out of date also, I can't see that being an issue 🤔

dan-mba:
> I don't think we have any pages that need near realtime data, so if other pages are 15mins out of date also, I can't see that being an issue 🤔

The problem that I was foreseeing assumes that the container is used in production.

After the container is built, it will sit in the registry until something happens in production that requires a pull of the container. I don't know how most applications handle rollout of an updated container, as I have no experience with it.

The build in the container could be minutes / hours / days old when it gets pulled to run in production.

The outdated page should get rebuilt after the first time it is served, because it is out of date.
This shouldn't present much of a problem, but it is something to think about.

Here is a link to documentation about sharing the next cache across pods using a shared mount when we migrate to that solution.
https://nextjs.org/docs/basic-features/data-fetching/incremental-static-regeneration#self-hosting-isr

Pradumnasaraf:
> What would the penalty be of creating a script for the CMD section of the Dockerfile that runs next build then next start?

This can be painful because this means every time a container will run it will rebuild the app because in `CMD` command is get triggered when the container run and in general practice we build things prior.

Pradumnasaraf:
> What would the penalty be of creating a script for the CMD section of the Dockerfile that runs next build then next start?

 @dan-mba This can be painful because this means every time a container will run it will rebuild the app because in `CMD` command is get triggered when the container run and in general practice we build things prior. What are your thoughts?

dan-mba:
> > What would the penalty be of creating a script for the CMD section of the Dockerfile that runs next build then next start?
> 
> @dan-mba This can be painful because this means every time a container will run it will rebuild the app because in `CMD` command is get triggered when the container run and in general practice we build things prior. What are your thoughts?

@Pradumnasaraf 
Does K8s provide a method to use a URL to test if a pod is ready for traffic?
If so, we could use this method and the on-demand re-validation to force the pages to re-validate after the server comes up.
https://nextjs.org/docs/basic-features/data-fetching/incremental-static-regeneration#using-on-demand-revalidation

That way the out of date pages would never get served to users.
That would be much better than forcing a rebuild every time a container starts.

Pradumnasaraf:
@dan-mba 
> Does K8s provide a method to use a URL to test if a pod is ready for traffic?
If so, we could use this method and the on-demand re-validation to force the pages to re-validate after the server comes up.
https://nextjs.org/docs/basic-features/data-fetching/incremental-static-regeneration#using-on-demand-revalidation

Yes, I did a quick research, there is a feature in k8s called `Readiness Probes`. This helps to determine whether a pod is ready to serve traffic or not. You can read more about it here: https://komodor.com/blog/kubernetes-health-checks-everything-you-need-to-know/

dan-mba:
> @dan-mba
> 
> > Does K8s provide a method to use a URL to test if a pod is ready for traffic?
> > If so, we could use this method and the on-demand re-validation to force the pages to re-validate after the server comes up.
> > https://nextjs.org/docs/basic-features/data-fetching/incremental-static-regeneration#using-on-demand-revalidation
> 
> Yes, I did a quick research, there is a feature in k8s called `Readiness Probes`. This helps to determine whether a pod is ready to serve traffic or not. You can read more about it here: https://komodor.com/blog/kubernetes-health-checks-everything-you-need-to-know/

@Pradumnasaraf 
Thanks for looking that up. It looks like a good solution when we start using K8s.
I will build out the re-validation API so it will be ready to go.

We will also need to create a separate Dockerfile for the deploy action.
This will handle pulling in the value for LINKFREE_MONGO_CONNECTION_STRING for the build.
Does passing it in as a build-arg from the deploy action make the most sense?

Pradumnasaraf:
> We will also need to create a separate Dockerfile for the deploy action.
This will handle pulling in the value for LINKFREE_MONGO_CONNECTION_STRING for the build.
Does passing it in as a build-arg from the deploy action make the most sense?

Yes, I think so. Having a single Dockerfile is always better, but not a problem if we have another one.

Pradumnasaraf:
> We will also need to create a separate Dockerfile for the deploy action.
This will handle pulling in the value for LINKFREE_MONGO_CONNECTION_STRING for the build.
Does passing it in as a build-arg from the deploy action make the most sense?

@dan-mba Yes, I think so. Having a single Dockerfile is always better, but not a problem if we have another one.

dan-mba:
> @dan-mba Yes, I think so. Having a single Dockerfile is always better, but not a problem if we have another one.

@Pradumnasaraf 
Can you review the changes I made to `deploy.yml` and the dockerfile I created. I used args to pass the url into the build step and export the environment variable on the run step for build to avoid it getting baked into the layer.

.github/workflows/deploy.yml
```diff
@@ -49,10 +49,11 @@ jobs:
         uses: docker/build-push-action@v3
         with:
           context: .
+          file: Dockerfile.deploy
           push: true
           secrets: |
             'GH_TOKEN=${{ secrets.GITHUB_TOKEN }}'
-          build-args: "github_token=${{ secrets.GITHUB_TOKEN }}"
+          build-args: "github_token=${{ secrets.GITHUB_TOKEN }},MONGODB_URL=${{ secrets.MONGODB_URL }}"
```
Pradumnasaraf, line 56: One of the great use cases of `build-args:`. Thanks learned something new :)

Dockerfile.deploy
```diff
@@ -0,0 +1,16 @@
+FROM node:18
+LABEL org.opencontainers.image.source https://github.com/eddiehubcommunity/LinkFree
+
+WORKDIR /usr/src/app
+
+COPY package*.json ./
+ARG NODE_ENV
+RUN if [ "$NODE_ENV" = "development" ]; then npm install --ignore-scripts; else npm install  --omit=dev --ignore-scripts; fi
```
Pradumnasaraf, line 8: We can remove the conditional, because we will not run the file via Compose

```suggestion
RUN npm install  --omit=dev --ignore-scripts
```

Pradumnasaraf (commented):
This is something amazing. Looks great, Dan. A small optimization.

dan-mba:
@Pradumnasaraf 
I updated the dockerfile to remove the node_env check as you suggested

@eddiejaoude 
Can you please add a secret to the Linkfree repo calle `MONGODB_UR` with the connection string to the production database in order for the deploy to work.

eddiejaoude:
OMG I missed this notification from last month Dan, so sorry!! Please tag me on Discord in the future for something urgent/blocking.

> Can you please add a secret to the Linkfree repo called MONGODB_URL with the connection string to the production database in order for the deploy to work.

Could we use the same name we have in the app itself `LINKFREE_MONGO_CONNECTION_STRING`?

dan-mba:
> Could we use the same name we have in the app itself `LINKFREE_MONGO_CONNECTION_STRING`?

Sure, I will change it.

eddiejaoude:
> > Could we use the same name we have in the app itself `LINKFREE_MONGO_CONNECTION_STRING`?
> 
> Sure, I will change it.

Thank you 👍 I added the GitHub secret

dan-mba:
> 
> Q. btw how would this work with Vercel?

In the Vercel workflow, you do a `vercel pull`.
This adds all the environment variables to a file used by `vercel build`.
So this should build with no problem in that workflow.

Pradumnasaraf (approved):
Looks good to merge.

eddiejaoude:
Silly question, how can test it locally?

dan-mba:
The easiest way is to compare the current version with my PR.

- add a `console.log()` to `getServerSideProps` or `getStaticProps`
- `npm run build` then `npm start`
- visit the index page & click refresh 5 times
- count the logged messages in the console

pages/api/revalidate.js
```diff
@@ -0,0 +1,17 @@
+export default async function handler(req, res) {
+  // Check for secret to confirm this is a valid request
+  if (
+    !req.query.secret ||
+    !process.env.MY_SECRET_TOKEN ||
+    req.query.secret !== process.env.MY_SECRET_TOKEN
```
eddiejaoude, line 6: Silly question but where do these secrets come from?

eddiejaoude (dismissed):
Great work 💪  this will really help with performance 🚀

pages/api/revalidate.js
eddiejaoude: Reading the docs because we have the time limit, I don't think we need the api to invalidate it? or am I missing something?

https://nextjs.org/docs/pages/building-your-application/data-fetching/incremental-static-regeneration#using-on-demand-revalidation

eddiejaoude: Ok thank you for explaining, I will update to use the existing secret we will use for the loading of the data `LINKFREE_API_SECRET`

Then we can merge

eddiejaoude (approved):
Looks great 👍  Other pages will benefit from this also 🚀

PR #6151, opened 2023-04-17
DRAFT: authentication bypass for github login test

amandamartin-dev:
Fixes Issue #5788 

## Check List (Check all the applicable boxes) 

- [ ] My code follows the code style of this project.
- [ ] My change requires changes to the documentation.
- [ ] I have updated the documentation accordingly.
- [ ] All new and existing tests passed.
- [ ] This PR does not contain plagiarized content.
- [ ] The title of my pull request is a short description of the requested changes.

## Note to reviewers

<a href="https://gitpod.io/#https://github.com/EddieHubCommunity/LinkFree/pull/6151"></a>

PR #6164, opened 2023-04-18
Test to mock GitHub oauth with Next-Auth

### LINKED ISSUE #5788: [OTHER] Playwright automated tests for authenticated pages

Playwright has an example of authenticated tests, to be able to test pages that require authentication

- [ ] Logged in, can reach Account page and check title
- [ ] Not logged in, trying to navigate to Account page, but redirected

Here are the Playwright docs https://playwright.dev/docs/auth but we don't want to use a real username and password, we want to mock those out, this blog might help https://dev.to/kuroski/writing-integration-tests-for-nextjs-next-auth-prisma-using-playwright-and-msw-388m

### Additional information

Plan B: We could during testing switch to credential login and create a fake user in the DB

eddiejaoude:
## Fixes Issue

Building on from https://github.com/EddieHubCommunity/LinkFree/pull/6151 

fixes #5788

- [x] save mock user to database
- [x] successfully log in with mock user

## Note to reviewers

<a href="https://gitpod.io/#https://github.com/EddieHubCommunity/LinkFree/pull/6164"></a>


amandamartin-dev:
I have a few theories I'm exploring as to why this won't work "as is" from the blog tutorial version. Feel free to follow my rabbit hole or consider others:

We may have to set up the mock server for this version to work even though we have no need to mock the API calls. I'm thinking this because the mock server is what intercepts and modifies the information allowing login to github. I got this idea re-reading this comment in the article where it talks about the data provided for the GH secrets and such won't matter since it will be intercepted. We are currently getting errors that it's invalid JWT becuase we are NOT intercepting it "// 3. Here we are binding our user with a "Github fake account", this is needed since we are using OAuth, we don't have to worry about this data since we are gonna intercept and mock the direct Github API calls"
Theory 2 -this won't work bc similarly to what Dan noticed, we need to provide more information, however, I'm not sure how to build a fake yet valid JWT. Another potential area to explore... (see [nextauth].js for the auth options that led me to this train of thought)
Still poking around but want to share my thoughts in case it shakes anything out of your head!

eddiejaoude:
Thanks Amanda! Another option/idea for plan C, is we enable email/password login during testing instead of GitHub OAuth, and we perform a real login but only to out DB (we would have to inject the user into the DB, and maybe have a password field that isn't used during production).

amandamartin-dev:
adding note to this issue from Discord suggestion to review https://next-auth.js.org/configuration/callbacks

amandamartin-dev:
> I did some debug work.
> 
> After modifying the tests & removing `storageState` which isn't really relevant for our use case, I found that next-auth doesn't like the `sessionToken` as it isn't a valid JWT.
> 
> There is a library to create valid [JWTs](https://github.com/auth0/node-jsonwebtoken) I will look into if using this will create a JWT that next-auth likes

Amazing! I was thinking the same thing after you noticed the JWT issue but hadn't looked into mocking that yet.  I'm about to head into some meetings so won't have time to look at this again until tomorrow likely so feel free to use this branch if you want. I haven't made any changes yet today

dan-mba:
I got the session Token working 🥳

Now the problem is the statistics page doesn't have all the info it needs in getServerSideProps.
Hopefully this is an easier problem to fix.

dan-mba:
The automated test fails because we nee `NEXTAUTH_SECRET` to be set in order for the test to run.

Not exactly sure how to do this.
Do we need to add a default setup to `.env.example`?

amandamartin-dev:
> The automated test fails because we nee `NEXTAUTH_SECRET` to be set in order for the test to run.
> 
> Not exactly sure how to do this. Do we need to add a default setup to `.env.example`?

I think you are right and from reading the docs, they may require it in the future anyway so we would be ahead of any future changes for prod and dev.

"Defining an explicit secret will make this problem go away. We will likely make this option mandatory, even in development, in the future."

Here's the docs for quick referencce: https://next-auth.js.org/configuration/options#secret

.env.example
```diff
@@ -5,4 +5,5 @@ NODE_ENV="development"
 
 GITHUB_ID=""
 GITHUB_SECRET=""
-NEXTAUTH_SECRET=""
+NEXTAUTH_SECRET="123456"
```
amandamartin-dev, line 8: Question for discussion (i do not know the answer) - can this value be anything or is it a better practice to generate it another way?  I know it's completely visible so I'm assuming it does not matter, but in the docs it's suggested to generate a "good" value like this: 

You can quickly create a good value on the command line via this openssl command.
```$ openssl rand -base64 32```

File: tests/setup/auth.js [hunk 2]
```diff
@@ -0,0 +1,100 @@
...
+        provider: "github",
+        providerAccountId: "2222222",
+        access_token: "ggg_zZl1pWIvKkf3UDynZ09zLvuyZsm1yC0YoRPt",
+        token_type: "bearer",
+        scope: "read:org,read:user,repo,user:email,test:all",
+      },
+      { new: true, upsert: true }
+    );
+  } catch (e) {
+    console.log(e, `failed for step 3`);
```
amandamartin-dev, line 75:
```suggestion
    console.error(e, `Test Account creation failed`);
```

.env.example
amandamartin-dev: ah, yes - you're correct. I missed that. Perfect.

dan-mba:
> Left a few suggestions, also another question I do not know the answer to. Is it fine to leave this test user sitting in the database? Will it be accidentally queried in any other code we have that we know of? We could destroy after tests. but then we are creating each time, which perhaps we don't want to do. Thoughts?

All tests are run on an empty database, not the production database.
All changes are discarded after the action runs.

amandamartin-dev:
> All tests are run on an empty database, not the production database. All changes are discarded after the action runs.

Oh! Thank you for that. It makes sense but just wasn't sure.  So I guess the only thing left is to officially open this to review? And then open a subsequent issue to write a test for the stats page (that addresses the data issues you noticed?)

dan-mba:
> Oh! Thank you for that. It makes sense but just wasn't sure. So I guess the only thing left is to officially open this to review? And then open a subsequent issue to write a test for the stats page (that addresses the data issues you noticed?)

I got rid of the data issue (we were passing session as a prop, but it was unused).

amandamartin-dev:
I can't switch the status of this PR out of draft, I think you have to @eddiejaoude but I"m going to open now and run the test once more with all of Dans changes to do a quick final pass

amandamartin-dev (approved):
ran in gitpod and checked the playwright reports for good measure LGTM!!

.env.example
eddiejaoude: yep in prod we use that command to generate a more secure secret

eddiejaoude:
Wow great work 🚀 ! I was out yesterday and looks like you both had fun and success 🎉

eddiejaoude:
Looks great, amazing work, thank you both! 

I kept getting timing outs on the search and map tests locally, and this was due to the use list api payload being so big, I hope you don't mind but I fixed it and added it to this PR - please review it though.

eddiejaoude:
I got too excited and merged it 😱 I am so excited we have this feature 🎉

PR #6416, opened 2023-04-28
chore: update docker image version

Pradumnasaraf:
## Changes proposed

- Version the `mongo` image in `mongo-compose.yml`, otherwise, users may have different versions of the image, which could lead to errors.

- Update the `docker-compose.yml` to 6.0.

<a href="https://gitpod.io/#https://github.com/EddieHubCommunity/LinkFree/pull/6416"></a>

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -21,7 +21,7 @@ services:
       - linkfree-network
 
   mongodb:
-    image: mongo:5.0
+    image: mongo:6.0
     container_name: mongodb
     networks:
       - linkfree-network
--- a/mongo-compose.yml
+++ b/mongo-compose.yml
@@ -2,6 +2,6 @@
 version: "3.9"
 services:
   mongodb:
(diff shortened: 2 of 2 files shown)


dan-mba:
@Pradumnasaraf Is there a specific reason for using 5.0 over 5?

I'm not sure how mongo uses semver, so do you believe it makes more sense to do testing before upgrading to a hypothetical 5.1.X if it is released?

Pradumnasaraf:
> @Pradumnasaraf We should be using 6.0 instead of 5.0. Atlas currently only supports 6.0

Hey, Dan

Oh, if this is the case, Iet's switch to 6.0 in `mongo-compose.yml`, as well as in `docker-compose.yml`.

dan-mba (approved):

eddiejaoude (approved):
Good spot and collaboration 👍 

yes we should be the same version at Atlas 💯

PR #6625, opened 2023-05-05
fix: remove clicks states (#6609)

### LINKED ISSUE #6609: [BUG] setState() is not get correct value
**update: code is no longer required and can be removed**

### Description

I found an issue in [components/user/UserLink.js](https://github.com/EddieHubCommunity/LinkFree/blob/main/components/user/UserLink.js) line number 29. Here we update state based on previous state, but current implementation of `setState()` is not correct. 
**Because**,
Whenever we update state and that depend on the previous state, for example, as we do it here [components/user/UserLink.js](https://github.com/EddieHubCommunity/LinkFree/blob/main/components/user/UserLink.js) line number 29. We should not do it like this but we should use an alternative form of this state updating function.

Screenshot (14)

### Why we shouldn't do it like this?
In many cases it will work fine, but according to React docs, React schedules state updates, it doesn't perform them instantly. And therefore, theoretically, if we schedule a lot of state updates at a same time, we could be depending on outdated or incorrect state snapshot if we use this approach.

### Screenshots

_No response_

### Additional information

Would love fix this if this a valid issue.
Thanks

badrivlog:
Closes #6609

## Changes proposed

clicks states is no longer required, as the clicks are not longer displayed on the profile page

## Note to reviewers

<a href="https://gitpod.io/#https://github.com/EddieHubCommunity/LinkFree/pull/6625"></a>

### DIFF
--- a/components/user/UserLink.js
+++ b/components/user/UserLink.js
@@ -1,11 +1,8 @@
-import { useState } from "react";
-
 import getIcon from "@components/Icon";
 import colors from "@config/icons.json";
 import Link from "@components/Link";
 
 export default function UserLink({ BASE_URL, link, username }) {
-  const [clicks, setClicks] = useState(link.clicks || 0);
   const DisplayIcon = getIcon(link.icon);
   let aria = "";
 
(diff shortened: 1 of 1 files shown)


Ananya2001-an (approved):
LGTM 👍🏼 Thanks for contributing 💯

badrivlog:
> LGTM 👍🏼 Thanks for contributing 💯 

My first contribution 👍 Thanks 💯💯💯

eddiejaoude (approved):
Looks great 👍 

Awesome to remove dead code 💪

badrivlog:
Thanks 💪

Issue #2893, opened 2023-01-08
Found two user profiles whose GitHub profiles does not exist


Aadarsh805:
### Description

1) `Skyhero-admin.json`
2) `aaquibdev.json`

this error occurred due to this `upstream image response failed for https://github.com/Skyhero-admin.png 404`

### Screenshots

this caused the random user section to generate only 4 user profiles, the 5th threw error
![Screenshot from 2023-01-09 00-40-14](https://user-images.githubusercontent.com/95094057/211214416-29c8c1e6-db39-4acf-926d-584dc6fc148c.png)

### Additional information

There must be more accounts like this, it might be negligible for a large number of users (this happened on my local dev server, with smaller user list)

eddiejaoude:
Thank you raising 👍

MD-ARBAB-SAHID:
> ### Description
> 1. `Skyhero-admin.json`
> 2. `aaquibdev.json`
> 
> this error occurred due to this `upstream image response failed for https://github.com/Skyhero-admin.png 404`
> 
> ### Screenshots
> this caused the random user section to generate only 4 user profiles, the 5th threw error ![Screenshot from 2023-01-09 00-40-14](https://user-images.githubusercontent.com/95094057/211214416-29c8c1e6-db39-4acf-926d-584dc6fc148c.png)
> 
> ### Additional information
> There must be more accounts like this, it might be negligible for a large number of users (this happened on my local dev server, with smaller user list)

@Aadarsh805 I also tried with dummy database with some users having GitHub profile and some not, but in that case it was using the fallbackSvg avatar instead of breaking from that point like as you mentioned because of unavailability 4th user 5th one was not displayed, I dont know if I'am getting it right or not 😅.

Aadarsh805:
> > ### Description
> > 
> > 1. `Skyhero-admin.json`
> > 2. `aaquibdev.json`
> > 
> > this error occurred due to this `upstream image response failed for https://github.com/Skyhero-admin.png 404`
> > ### Screenshots
> > this caused the random user section to generate only 4 user profiles, the 5th threw error ![Screenshot from 2023-01-09 00-40-14](https://user-images.githubusercontent.com/95094057/211214416-29c8c1e6-db39-4acf-926d-584dc6fc148c.png)
> > ### Additional information
> > There must be more accounts like this, it might be negligible for a large number of users (this happened on my local dev server, with smaller user list)
> 
[...]

PR #6668, opened 2023-05-07
fix: Changed the user 'Skyhero-admin' to 'Skyhero-alt'

Skyhero-alt:
## Fixes Issue
 Found two user profiles whose GitHub profiles does not exist #2893 

Updated `Skyhero-admin.json` and renamed it to `Skyhero-alt.json`
And updated the profile github link since the last account doesn't exist anymore

<a href="https://gitpod.io/#https://github.com/EddieHubCommunity/LinkFree/pull/6668"></a>


SaraJaoude:
The package lock file should only be committed if the package.json file has changes. There could be two reasons why the lock file has been updated for you:

1. You used `npm install` instead of `npm ci` or
2. You have GitHub Actions running on your fork. This should be disabled by default, please do not enable.

Please undo any changes to the package lock file. 

Here is a video that will show you how to do it: https://youtu.be/jHkIXYVpMTo

SaraJaoude:
@Skyhero-alt I can see that you have clicked on update branch, but this doesn't actually solve the changes to the package lock file.  

I will merge into a temporary branch and fix.

PR #6811, opened 2023-05-12
fix: External URLs in footer now open in new tab

### LINKED ISSUE #6763: GitHub links - Redirect to a new tab
### Description

The Github links which are present in the footer when clicked can be made to be opened in a new tab , instead of redirecting. The only Giithub link which opens in a new tab are the one which is present in the nav-bar icon and the user's Github profile.

### Screenshots

image

### Additional information

_No response_

HassanTanveer:
## Fixes Issue
- External links in footer are now opened in a new tab

closes #6763 

## Note to reviewers

<a href="https://gitpod.io/#https://github.com/EddieHubCommunity/LinkFree/pull/6811"></a>


components/Footer.js
```diff
@@ -11,6 +11,8 @@ import Button from "./Button";
 import Link from "@components/Link";
 
 export default function Footer() {
+  const isExternalUrl = (url) => /^(?:[a-z+]+:)?\/\//i.test(url)
```
kaafivikrant, line 14: It would be more readable if we add the regex in a const variable.

const EXTERNAL_URL_REGEX = /^(https?|ftp):\/\//i;

const isExternalLink = (url, regex = EXTERNAL_URL_REGEX) => regex.test(url);

Something like this, we can try.

HassanTanveer: Good suggestion, I'll try to improve readability.

eddiejaoude (changes requested):
I think just having a flag in the json is simpler and then we have more control.

For example

```js
{
        name: "YouTube",
        href: "https://www.youtube.com/watch?v=05HEeCQSKRE&list=PL4lTrYcDuAfyU0fJcCGLm5r-hM_rqXaxd",
        icon: FaYoutube,
        external: true
      },
```

HassanTanveer:
@eddiejaoude Yeah that was the first approach that came to mind but I ended up over-complicating the issue. I'll make changes according to your suggestion

eddiejaoude:
> @eddiejaoude Yeah that was the first approach that came to mind but I ended up over-complicating the issue. I'll make changes according to your suggestion

thank you for updating 👍 it is interesting to see this approach, but it wouldn't be consistent with other areas of the application - something like this would have to be an application wide solution, but as you mentioned it is probably over engineering it

eddiejaoude:
I did something similar for the Roadmap page https://github.com/EddieHubCommunity/LinkFree/pull/6822

HassanTanveer:
> > @eddiejaoude Yeah that was the first approach that came to mind but I ended up over-complicating the issue. I'll make changes according to your suggestion
> 
> thank you for updating 👍 it is interesting to see this approach, but it wouldn't be consistent with other areas of the application - something like this would have to be an application wide solution, but as you mentioned it is probably over engineering it

Of course, consistency throughout the code is important 👍🏻

iamthecloverly (approved):

eddiejaoude (approved):
Looks good 👍

PR #6836, opened 2023-05-13
fix: add useEffect dependency

### LINKED ISSUE #6693: [BUG] useEffect has missing dependency
### Description

found an issue in [pages/search.js](https://github.com/EddieHubCommunity/LinkFree/blob/main/pages/search.js) line number `120`  useEffect has missing dependency `filterData`.

According to [React Docs](https://react.dev/learn/lifecycle-of-reactive-effects#all-variables-declared-in-the-component-body-are-reactive), props, state and others values declared inside the component are reactive because they're calculated during rendering and participate in the React data flow. If useEffect uses any reactive values then that values must be included in the dependency array.

Screenshot (17)
[pages/search.js](https://github.com/EddieHubCommunity/LinkFree/blob/main/pages/search.js) in useEffect  body line number `116` we call `filterData` function but never include `filterData` as dependency, this seems to be not correct.

### Screenshots

_No response_

### Additional information

Would love to fix this issue.

Thanks

badrivlog:
* fix: add useEffect dependency
* add useCallback hook
* update pages/search.js

## Fixes Issue
Closes #6693 

## Changes proposed
add useEffect dependency

### DIFF
--- a/pages/search.js
+++ b/pages/search.js
@@ -1,4 +1,4 @@
-import { useEffect, useState, useRef } from "react";
+import { useEffect, useState, useCallback } from "react";
 import { useRouter } from "next/router";
 
 import UserCard from "@components/user/UserCard";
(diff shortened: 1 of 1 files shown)


pages/search.js
```diff
@@ -44,45 +44,47 @@ export default function Search({ data }) {
...
   useEffect(() => {
     if (username) {
       setNotFound(username);
     }
   }, [username]);
 
-  const filterData = (value) => {
-    const valueLower = value.toLowerCase();
-    const terms = valueLower.split(",");
+  const filterData = useCallback(
```
badrivlog, line 53: By wrapping `fitlerData` in `useCallback` hook, we ensure that it's the same function between re-renders(until dependency change).

pages/search.js
```diff
@@ -44,45 +44,47 @@ export default function Search({ data }) {
...
-    }
+      if (results.length) {
+        setNotFound();
+      }
 
-    setFilteredUsers(results.sort(() => Math.random() - 0.5));
-  };
+      setFilteredUsers(results.sort(() => Math.random() - 0.5));
+    },
+    [users]
```
badrivlog, line 86: `filterData` function re-created only when `users` value changed

badrivlog:
here the same code is added and subtracted, because when I wrap `filterData` in `useCallback` hook then prettier format  that code and creates this annoying behaviour. I have been struggling with this issue for a couple of days. Please give me feedback, I will be grateful to you.

gptkrsh (approved):
Looks good to me!

badrivlog:
Hey @gptkrsh , thanks so much for reviewing my PR 👍

PR #6885, opened 2023-05-15
Refined the mongo.js config file, Closes #6855

### LINKED ISSUE #6855: [OTHER] Fix mongo db initial connection config

Currently, to establish a connection with the mongoDB instance, we use the `connectMongo` function in the `config/mongo.js` file. A new connection is not established when there is an existing connection. Please refer to the attached screenshot:
Screenshot 2023-05-13 at 11 28 14 PM

`hasConnection` variable is assigned in line 24 and reassigned in line 28. 

Proposal:

1. replace `hasConnection` with `connection` variable which stores the result of `await mongoose.connect(....` which returns an object with the current connection's information.
2. Remove the return and re assignment of `hasConnection` since we don't use the connection information anywhere.
3. Improvement: extract `LINKFREE_MONGO_CONNECTION_STRING` variable using `const { } = params.env` and re use it wherever needed. This way, we don't have to use `process.env.....` everywhere.
4. Remove the usage of `process.env.CA_CERT` as it is not being used anymore.
5. Fix `import logger from "../config/logger";` to `import logger from "./logger";` since the `logger.js` file exists in the same directory.

### Additional information

_No response_

Chalithya:
## Fixes Issue

Closes #6855

## Changes proposed

1. replace hasConnection with connection variable which stores the result of await mongoose.connect(.... which returns an object with the current connection's information.
2.Remove the return and re assignment of hasConnection since we don't use the connection information anywhere.
3.Improvement: extract LINKFREE_MONGO_CONNECTION_STRING variable using const { } = params.env and re use it wherever needed. This way, we don't have to use process.env..... everywhere.
4.Remove the usage of process.env.CA_CERT as it is not being used anymore.
5.Fix import logger from "../config/logger"; to import logger from "./logger"; since the logger.js file exists in the same directory.

## Note to reviewers

I removed returns inside try catch as it returns a Promise and throwed an error inside the catch block to reduce the number of lines in the code. 

<a href="https://gitpod.io/#https://github.com/EddieHubCommunity/LinkFree/pull/6885"></a>


config/mongo.js
```diff
@@ -1,36 +1,29 @@
...
-    );
-    hasConnection = true;
-    logger.info("MongoDB connected");
-    return hasConnection;
-  } catch (e) {
-    hasConnection = false;
-    logger.error(e, "DB connection failed");
+  try {
+    const options = { autoIndex: true, family: 4 };
+    connection = await mongoose.connect(LINKFREE_MONGO_CONNECTION_STRING,options);
```
mnvsk97, line 22:
```suggestion
    connection = await mongoose.connect(LINKFREE_MONGO_CONNECTION_STRING, options);
```

config/mongo.js
```diff
@@ -1,36 +1,29 @@
...
+    logger.info("MongoDB connected:", connection.name);
```
mnvsk97, line 23: Hey 👋 we can also improve this log to something like 
`DB connection successful:` to keep it in sync with the error log in line 25.

config/mongo.js
```diff
@@ -1,36 +1,29 @@
...
+
+  if (!LINKFREE_MONGO_CONNECTION_STRING) {
     throw new Error(
       "Please define the LINKFREE_MONGO_CONNECTION_STRING environment variable (if local add to .env file)"
     );
   }
 
-  if (hasConnection) {
-    return hasConnection;
+  if (connection) {
```
mnvsk97, line 16: This check to see if the connection exists should be at the start of the function. If a connection exists, the function can return from there itself.

mnvsk97:
Hey [Chalithya](https://github.com/Chalithya) 👋 
I've added a few comments. Please have a look and let me know if you me to elaborate on any of the comments. Thanks for working on this issue !

config/mongo.js
Chalithya: Yeah. It's true. We don't need to check connection strings if there is existing connection available

eddiejaoude:
Great collaboration 💪

config/mongo.js
```diff
@@ -1,36 +1,29 @@
 import mongoose from "mongoose";
-import * as fs from "fs";
+import { env } from "process";
```
eddiejaoude, line 2: I don't think this is needed

config/mongo.js
```diff
@@ -1,36 +1,29 @@
...
-    );
-    hasConnection = true;
-    logger.info("MongoDB connected");
-    return hasConnection;
-  } catch (e) {
-    hasConnection = false;
-    logger.error(e, "DB connection failed");
+    const options = { autoIndex: true, family: 4 };
+    connection = await mongoose.connect(LINKFREE_MONGO_CONNECTION_STRING, options);
+    logger.info("DB connection successful:", connection.name);
```
eddiejaoude, line 23: I think this is missing a `return connection` otherwise it won't work on the first call when there is no connection

eddiejaoude (changes requested):
Looks good 👍 

Great collaboration! I left some small comments

eddiejaoude:
I will merge and tweak in a temporary branch

PR #7138, opened 2023-05-23
data:MehmetBaloglu

Honey171:


amandamartin-dev (changes requested):
Hello! You are close but need a few adjustments.
your file must be in the data folder and your file name MUST exactly match your github user name so you should move the file and rename so that is looks like this data/Honey171.json

Honey171:
I assume like this right?

amandamartin-dev (approved):

PR #7191, opened 2023-05-26
Create script to load database

dan-mba:
## Changes proposed
Since the change to using database requests for the app, the development process now requires populating the database.
This PR will provide a script to handle that.
It will also do that automatically in `docker-compose` and Gitpod


config/mongo.js
```diff
@@ -1,5 +1,5 @@
 import mongoose from "mongoose";
-import logger from "./logger";
```
eddiejaoude, line 2: Is this extension needed?

dan-mba: Yes, it is one of many annoying things required to use es6 modules in node.

eddiejaoude:
Wow Dan you put so much effort into this PR! But I think it has too many changes when the simpler solution would be for the dev to visit the url ...

- The info and url is in the docs and README
- Also there is a PR for when in dev mode to load data, an alert is shown to prompt the dev   https://github.com/EddieHubCommunity/LinkFree/pull/7240

Sorry, I will close this PR.

PR #7240, opened 2023-05-28
Reload json data alert when in dev mode

### LINKED ISSUE #7020: [FEATURE] In development have alert with button to load data
### Description

Now the json files are loaded into the DB during deployment, for local development we need to load data into the DB otherwise the app is empty

To load data we can visit a url http://localhost:3000/api/system/reload?secret=development

TODO
- only appears in development mode and not on vercel (2 checks)
- message and button to hit API and reload data
- add to layout to appear on every page
- do not use secret variable because might leak data if every display on prod, hardcode to `development`

### Screenshots

_No response_

### Additional information

_No response_

eddiejaoude:
## Fixes Issue

fixes #7020

Screenshot 2023-05-28 at 20 37 48

## Note to reviewers

<a href="https://gitpod.io/#https://github.com/EddieHubCommunity/LinkFree/pull/7240"></a>

### DIFF
--- a/components/Alert.js
+++ b/components/Alert.js
@@ -1,8 +1,9 @@
 import { FaInfoCircle } from "react-icons/fa";
 import { MdError, MdOutlineWarningAmber } from "react-icons/md";
 import { FaRegCheckCircle } from "react-icons/fa";
+import Button from "./Button";
 
-export default function alert({ type, message }) {
+export default function Alert({ type, message, url, action }) {
   switch (type) {
     case "success":
       return (
@@ -11,7 +12,8 @@ export default function alert({ type, message }) {
           role="alert"
         >
           <FaRegCheckCircle className="mr-2" />
-          <span>{message}</span>
+          <span className="grow">{message}</span>
+          {url && <Button href={url} text={action} primary={true} />}
         </div>
       );
       break;
@@ -22,7 +24,8 @@ export default function alert({ type, message }) {
           role="alert"
         >
           <MdError className="mr-2" />
-          <span>{message}</span>
+          <span className="grow">{message}</span>
+          {url && <Button href={url} text={action} primary={true} />}
         </div>
       );
       break;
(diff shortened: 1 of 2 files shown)


dan-mba:
I am not a fan of the developer experience on this.
The alert will always be shown on all pages in development & there is no way to turn it off outside of setting the environment variable.

Another issue is what setting does `NEXT_PUBLIC_VERCEL_ENV` have in the GitHub action build?
If this is not `production` there it will cause huge problems.

eddiejaoude:
- it could be a notification rather than an alert box?
- I could set the environment variable in the GitHub action and/or do another check?

dan-mba:
I put together some ideas on how to stop displaying the message after the data was loaded.
I will put it together in a separate branch in the morning.

eddiejaoude:
> I put together some ideas on how to stop displaying the message after the data was loaded.
> I will put it together in a separate branch in the morning.

Sure. The alert could check for users in the db on an existing endpoint like `/api/statistics/totals` 

```js
{
   "views": 3054,
   "clicks": 40,
   "users": 1962,
   "active": 1990
}
```

dan-mba:
I implemented the code to disable the alert after load in a branch here:
https://github.com/EddieHubCommunity/LinkFree/tree/alerts-check

eddiejaoude:
> I implemented the code to disable the alert after load in a branch here:
> https://github.com/EddieHubCommunity/LinkFree/tree/alerts-check

Thank you 👍 

What did you think of my idea to use `/api/statistics/totals` and the `users` property from the response?

dan-mba:
> > I implemented the code to disable the alert after load in a branch here:
> > https://github.com/EddieHubCommunity/LinkFree/tree/alerts-check
> 
> Thank you 👍
> 
> What did you think of my idea to use `/api/statistics/totals` and the `users` property from the response?

There is no way to call the API from where you added the alert component.
I added a prop which I set in getServerSideProps.
FYI, the users count from the totals API is always greater than 0. It still checks the files in the data directory.

eddiejaoude:
> There is no way to call the API from where you added the alert component. 

oh that could be a problem 

> FYI, the users count from the totals API is always greater than 0. It still checks the files in the data directory.

oh really? I thought that was updated to read the db

eddiejaoude:
I like your idea of doing data checks on each page 👍 

https://github.com/EddieHubCommunity/LinkFree/compare/main...alerts-check

PR #7356, opened 2023-06-03
Set preload to false on most links

dan-mba:
## Changes proposed
The additional JavaScript loaded by preload causes performance issues on mobile.
The NextJS configuration of Webpack will not create a separate main chunk for links that are preloaded on desktop only.
This PR will set all links using our Link component to `preload={false}`.

<a href="https://gitpod.io/#https://github.com/EddieHubCommunity/LinkFree/pull/7356"></a>


components/Link.js
```diff
@@ -9,6 +9,7 @@ export default function Link({ children, className, rel, ...restProps }) {
           ? className
           : "text-blue-600 underline decoration-dotted dark:text-blue-500 hover:underline hover:decoration-solid"
       }
+      prefetch={false}
```
eddiejaoude, line 12: Do we want to have a condition so it only happens on mobile?

dan-mba: I tried that, but webpack still bundles the code to preload the JavaScript files if the preload is set by a conditional. It is around 700k of JS that gets loaded by the main-XXX chunk of the bundle without preload set to false.

eddiejaoude: ouch, ok thank you for clarifying 👍

eddiejaoude (approved):
Looks good 👍

PR #7363, opened 2023-06-04
fix: vercel single deploy

eddiejaoude:
## Fixes Issue

Vercel GH Action deploys when merged to `main` and also when the release is created (so 2 deploys that are the same)

- we want to deploy on release
- but we also want to deploy when data files are merged to `main` because no release us created
- this will only deploy on `main` for data files

Screenshot 2023-06-04 at 07 41 16

From GitHub docs https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#running-your-workflow-only-when-a-push-to-specific-branches-occurs

Screenshot 2023-06-04 at 07 43 58

### DIFF
--- a/.github/workflows/vercel.yml
+++ b/.github/workflows/vercel.yml
@@ -7,8 +7,12 @@ on:
   release:
     types: [published]
   push:
+    # this will deploy data files only when no release is created
     branches:
-      - main # this will deploy data files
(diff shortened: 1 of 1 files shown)


dan-mba (approved):
Looks good to me

PR #7379, opened 2023-06-04
Run profile stat updates concurrently

dan-mba:
## Changes proposed
For each profile view, we perform 4 separate stat updates.
Each of these performs a write to MongoDB.
We currently wait for each operation to complete before performing the next.
Running these operations concurrently instead of one after another will improve performance.
Because these operation are all writes & involve network connections, there is a delay.
Node does not allow running in parallel, but will perform work on another task while one is waiting for results.

<a href="https://gitpod.io/#https://github.com/EddieHubCommunity/LinkFree/pull/7379"></a>


olasunkanmi-SE (commented):
I think it will be cleaner if you define each of the db opearions in a seperate functions. This gives a clear understanding of what each of the code blocks does. For example
` const updateDate = async (date, username) =>{
  try {
      await Stats.updateOne(
        {
          date,
        },
        {
          $inc: { users: 1 },
        },
        { upsert: true }
      );
      log.info(`app profile stats incremented for username: ${username}`);
    } catch (e) {
      log.error(e, `app profile stats failed for ${username}`);
    }
}`
Then you can promise.all([updateDate, ..., ...])

pages/api/users/[username]/index.js
```diff
@@ -107,79 +107,90 @@ export async function getUserApi(req, res, username) {
...
-        }
-      );
-      log.info(`stats incremented for username: ${username}`);
-    } catch (e) {
-      log.error(
-        e,
-        `failed to increment profile stats for username: ${username}`
-      );
-    }
+    updates.push((async () => {
```
Comment (eddiejaoude) [inline pages/api/users/[username]/index.js:115]: I think this could be done without the function and await because you have the promise all?

Comment (dan-mba) [reply pages/api/users/[username]/index.js:115]: You need to create an async function to return a promise. Then add that promise to the array. Promise.allSettled collects all the promises in the array into a single promise that will resolve/reject once all promises are settled. Then you await that single promise.

eddiejaoude (approved):
Looks good 👍

eddiejaoude:
Sorry the path got renamed in another PR from `users` to `profiles` which caused a conflict

pages/api/users/[username]/index.js
Comment (eddiejaoude) [reply pages/api/users/[username]/index.js:115]: ok thank you 👍

PR #7535, opened 2023-06-12
feat: search by location

### LINKED ISSUE #3467: [FEATURE] Adding Search by Location capability
### Description

It would be great if we can search for users based on their location. It will help us in connecting with like-minded persons around us.

If we could add location information to our profile in [CITY], [STATE], [COUNTRY]. format, then it can be integrated with the search engine.

Examples:
- Kolkata, West Bengal, India.
- Mumbai, Maharashtra, India.
- Capital District, New York, United States.

### Screenshots

image

### Additional information

_No response_

shoebilyas123:
## Fixes Issue
Closes #3467

## Changes proposed

- Added city, state and country fields to profile schema
- Added search by location feature on search page

## Note to reviewers

<a href="https://gitpod.io/#https://github.com/EddieHubCommunity/LinkFree/pull/7535"></a>

### DIFF
--- a/models/Profile.js
+++ b/models/Profile.js
@@ -23,6 +23,9 @@ const ProfileSchema = new mongoose.Schema(
       lat: Number,
       lon: Number,
       updatedAt: Date,
+      city: String,
+      state: String,
+      country: String,
     },
     views: {
       type: Number,
--- a/models/Profile.js
+++ b/models/Profile.js
@@ -23,9 +23,6 @@ const ProfileSchema = new mongoose.Schema(
       lat: Number,
       lon: Number,
       updatedAt: Date,
-      city: String,
-      state: String,
-      country: String,
     },
     views: {
       type: Number,
--- a/pages/search.js
+++ b/pages/search.js
@@ -20,7 +20,7 @@ export async function getStaticProps() {
     tags: [],
   };
   try {
-    data.users = await getUsers({ });
+    data.users = await getUsers();
   } catch (e) {
     logger.error(e, "ERROR search users");
   }
--- a/pages/search.js
+++ b/pages/search.js
@@ -1,6 +1,5 @@
 import { useEffect, useState } from "react";
 import { useRouter } from "next/router";
-import { pick } from "lodash";
 
 import UserCard from "@components/user/UserCard";
 import Alert from "@components/Alert";
@@ -21,7 +20,7 @@ export async function getStaticProps() {
     tags: [],
   };
   try {
-    data.users = await getUsers({ cards: false });
+    data.users = await getUsers({ });
   } catch (e) {
     logger.error(e, "ERROR search users");
   }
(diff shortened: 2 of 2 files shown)


models/Profile.js
```diff
@@ -23,6 +23,9 @@ const ProfileSchema = new mongoose.Schema(
       lat: Number,
       lon: Number,
       updatedAt: Date,
+      city: String,
```
dan-mba, line 26: There is no reason to modify the database model as you are not writing to the database.

pages/search.js
```diff
@@ -1,5 +1,6 @@
 import { useEffect, useState } from "react";
 import { useRouter } from "next/router";
+import { pick } from "lodash";
```
dan-mba, line 3: Lodash is not a direct dependency & there is probably a way to do the same thing without including it.

pages/search.js
```diff
@@ -20,7 +21,7 @@ export async function getStaticProps() {
     tags: [],
   };
   try {
-    data.users = await getUsers({ cards: true });
+    data.users = await getUsers({ cards: false });
```
dan-mba, line 24: there is no need to include the object as a parameter as cads false is the default.

models/Profile.js
shoebilyas123: But according the discussion,  extra fields including city, state and country were suggested to be added in the profile. That's why I modified the schema.

dan-mba: While this may have been the original idea, #3169 implemented location by pulling the data from GitHub, then obtaining the coordinates & full name from a mapping system.

Changing this would require more discussion outside the scope of using the existing location data for search.

pages/search.js
eddiejaoude: Good spot, yes we don't want to include `lodash`

pages/search.js
```diff
@@ -85,19 +84,23 @@ export default function Search({ data: { users, tags, randUsers } }) {
       const nameLower = user.name.toLowerCase();
       const usernameLower = user.username.toLowerCase();
       const userTagsString = user.tags.join(", ").toLowerCase();
-      
+      const userLocationString = [user.location.provided]
```
dan-mba, line 87: This should be `const userLocationString = user.location ? user.location.name.toLowerCase() : "";`

dan-mba (changes requested):

pages/search.js
shoebilyas123: Fixed

pages/search.js
```diff
@@ -85,19 +84,23 @@ export default function Search({ data: { users, tags, randUsers } }) {
       const nameLower = user.name.toLowerCase();
       const usernameLower = user.username.toLowerCase();
       const userTagsString = user.tags.join(", ").toLowerCase();
-      
+      const userLocationString = user.location ? [user.location.provided]
```
dan-mba, line 87: There is no reason to create an array then perform a join on it.
This combination essentially does nothing, because the array only has one element.

dan-mba (changes requested):

pages/search.js
dan-mba: While there is a possibility the location information will change in the future, there is no benefit to adding additional complexity now that will still need to be modified in the future.

shoebilyas123: Sure! I have fixed it.

dan-mba (approved):
Looks good to me

eddiejaoude:
Great collaboration 💪

PR #7705, opened 2023-06-20
Improve linting

dan-mba:
## Changes proposed
Configure eslint to check for more issues.
Fix issues found by new eslint configuration


.eslintrc.json
```diff
@@ -1,3 +1,29 @@
 {
-  "extends": ["next/core-web-vitals", "plugin:storybook/recommended"]
+  "env": {
+    "browser": true,
+    "es6": true,
+    "node": true
+  },
```
krishstwt, line 6: This config doesn't include any style-enforcing configs, do you think we should have something like that too? Otherwise formatting will have to still go to eslint, I suggest `standardjs` convention

dan-mba: Adding style conventions to eslint will make it significantly slower & cause build breakage.

I think most styling could be handled by prettier.
We just need to better integrate prettier.
Currently it is only used to format the data JSON files.

Issue #7349, opened 2023-06-03
[FEATURE] Rename GitHub user


SaraJaoude:
### Description

- After the issue https://github.com/EddieHubCommunity/LinkFree/issues/7065 is done
- perform check and display alert if logged in username does not matched associated profile
- have action button to renamed all their data to match their new username
- show warning that their url will change from X to Y, but they will keep their stats
- they need to type their username to confirm

xkrishguptaa:
What is the update on this one? Really interested for this one!

theunhackable:
@xkrishguptaa step 1 is done so I think you can ask for this issue to be assigned

xkrishguptaa:
I don't believe I have right now enough time to do this due to exams, but who so ever picks it up, I'm happy to guide them and help them if they face any errors.

sidrasa:
I want to work on this issue.

SaraJaoude:
> I want to work on this issue.

@sidrasa I have assigned this to you.

SaraJaoude:
@sidrasa as it has been two weeks since this was assigned to you and as mentioned in our [Contributing Guide ](https://github.com/EddieHubCommunity/BioDrop/blob/main/CONTRIBUTING.md) since no Pull Request has been created, I have unassigned this from you to give someone else a chance to work on this issue.

akash19coder:
Isn't it the case that issue #8435 simiar to the one we are talking about here?

eddiejaoude:
Thank you @akash19coder, yes it is 👍 I will close https://github.com/EddieHubCommunity/BioDrop/issues/8435 because it is a duplicate

Vijaykv5:
Please close this issue @eddiejaoude @SaraJaoude as it's already fixed

eddiejaoude:
thank you @Vijaykv5 but I don't think this has been fixed, but we are not going to do it, so I will close it for now

PR #7721, opened 2023-06-21
Issue: #7065 [FEATURE] Associate profile to logged in user

### LINKED ISSUE #7065: [FEATURE] Associate profile to logged in user
### Description

>  if username is different from logged in user, show a warning and have a button to update (this will only work if someone logged in before they changed their username)

USER
- [ ] When user logs in associate the profile to their account using their GitHub username (maybe GitHub ID?)

Ideas...
- I think the model `models/Account.js` needs to be updated to create a on-to-many relationship to `models/Profile.js` on login (or on profile update in PR #6998?) 
- then the queries in the app should not check `Profile` model for username but use the relationship of the logged in user
- this will then help us resolve the issue of people changing their GitHub username and also allow people to create multiple profiles (in the future)

These 2 have been moved to their own issue https://github.com/EddieHubCommunity/LinkFree/issues/7349
- [ ] ~~If the username no longer matches the logged in user, show an error of their previous associated username and their current GitHub username~~
- [ ] ~~When displaying the error, have a button to update their username and their url will also change~~

~~Dependent on https://github.com/EddieHubCommunity/LinkFree/issues/7066~~

### Screenshots

_No response_

### Additional information

After this issue is complete, then this issue needs to be done https://github.com/EddieHubCommunity/LinkFree/issues/7349

raviraj-github:
## Fixes Issue

Closes #7065

## Changes proposed
- [x]  - The account model has been extended to facilitate a one-to-many relationship with profiles.
- [x]  - An account API has been implemented to retrieve account information from the database and establish profile associations during the initial user login.
- [x]  - Statistics.js has been updated to call the account API.

## Screenshots

image

## Note to reviewers

I have used info log for account js but that could be changed to debug once the logging issue PR merged. 

<a href="https://gitpod.io/#https://github.com/EddieHubCommunity/LinkFree/pull/7721"></a>


pages/account/statistics.js
```diff
@@ -51,6 +57,11 @@ export async function getServerSideProps(context) {
     };
   }
 
+  //Linking a profile to an account is a one-time process. Subsequent logins will disregard any attempt to add the profile again.  
+  if (account) {
```
eddiejaoude, line 61: I think this condition will always be true, because the login would fail

pages/api/account/account.js
```diff
@@ -0,0 +1,72 @@
...
+  if (!session) {
+    res.status(401).json({ message: "You must be logged in." });
+    return;
+  }
+
+  if (req.method !== "GET") {
+    return res.status(400).json({ error: "Invalid request: GET request required" });
+  }
+
+  res.status(200);
```
eddiejaoude, line 17: Should this call the other function in the file?

pages/api/account/account.js
```diff
@@ -0,0 +1,72 @@
...
+    if (existingProfile) {
+      logger.info('Profile already exists in the account. Skipping addition.' + JSON.stringify(account));
+      return account;;
+    }
+
+    // Add the newProfile to the profiles array
+    account.profiles.push(newProfileId);
+
+    // Save the updated account
+    await account.save();
```
eddiejaoude, line 61: Can this be done in Mongo, something like...

```js
   await Account.findOneAndUpdate(
      {
        _id: providerAccountId,
        "profile._id": new ObjectId(id),
      },
      {
        $set: {
          "profiles.$": newProfile,
        },
      },
      { upsert: true, new: true }
    );
```

eddiejaoude (changes requested):
Looks good 👍 

I left a few inline comments. Also maybe the naming is too long, even though it is descriptive

pages/api/account/account.js
raviraj-github: Yes, It should call getAccountByProviderAccountId and return the account.

pages/account/statistics.js
raviraj-github: You are saying the GitHub login will never fail right? I have added this just for the case where we will have our own signup and login page. But can be ignored for now.

pages/api/account/account.js
raviraj-github: I have already made a call for an account once from DB. And I think this function "findOneAndUpdate" will load the account one more time. That's the reason have passed the account from statistics.js. But I will check once.

pages/account/statistics.js
raviraj-github: For now, let's remove if condition and call the method directly. Maybe in the future, we will see.

eddiejaoude: > You are saying the GitHub login will never fail right

I mean if it fails, it would not get to this page, `next-auth` would redirect somewhere else before reach this code

raviraj-github: Ahaa. That I will take care in another issue https://github.com/EddieHubCommunity/LinkFree/issues/7349. Where the actual call be made. 

image


See the highlighted code.

pages/api/account/account.js
raviraj-github: I have examined both `updateOne` and `findOneAndUpdate`, and they both involve making a database call to perform updates. In my code, I verify the profile mappings by checking the previously loaded account in `statistics.js`. Only if there is no match found, I make a database call to save the profile. This approach helps optimize the process by minimizing unnecessary database operations.

raviraj-github:
Are you talking about the function name here?

raviraj-github:
I will create a new PR with all the changes. Thanks

pages/account/statistics.js
raviraj-github: Closed this comment in PR https://github.com/EddieHubCommunity/LinkFree/pull/8232

pages/api/account/account.js
raviraj-github: No changes for this in PR  https://github.com/EddieHubCommunity/LinkFree/pull/8232 comment as I have explained in the last comment.

PR #7782, opened 2023-06-25
Manage forms - pages to tailwind config colors

sandramsc:
## Change color assignments to **tailwind.config.js** colors:

Discussed in live stream and GitHub Discussion https://github.com/EddieHubCommunity/LinkFree/discussions/7770

- `text-gray-400` (any color or value after the color is incorrect way and should be using text-primary-medium )

- same with` bg-white` (using the color is incorrect)

- the available colors in our project are here in the **tailwind config** under colors https://github.com/EddieHubCommunity/LinkFree/blob/main/tailwind.config.js 

-  we should not use any of these text-[color] or text-[color]-[value] or with bg also like bg-[color] or bg-[color]-[value]

-------------------------------------------------------------------------------------

**Pages changed**:
username.js
changelog.js
index.js
newsletter.js
playground.js
premium.js
roadmap.js
terms.js

<a href="https://gitpod.io/#https://github.com/EddieHubCommunity/LinkFree/pull/7782"></a>

PR #7791, opened 2023-06-25
Manage forms - files to tailwind config colors

sandramsc:
## Change color assignments to **tailwind.config.js** colors:

Discussed in live stream and GitHub Discussion https://github.com/EddieHubCommunity/LinkFree/discussions/7770

- `text-gray-400` (any color or value after the color is incorrect way and should be using text-primary-medium )

- same with` bg-white` (using the color is incorrect)

- the available colors in our project are here in the **tailwind config** under colors https://github.com/EddieHubCommunity/LinkFree/blob/main/tailwind.config.js 

-  we should not use any of these text-[color] or text-[color]-[value] or with bg also like bg-[color] or bg-[color]-[value]

-------------------------------------------------------------------------------------
**Continued with color updates in files**.

**Pages fixed from 1st related PR**:
username.js

Before: icon

After: icon

<a href="https://gitpod.io/#https://github.com/EddieHubCommunity/LinkFree/pull/7791"></a>


eddiejaoude:
Thanks @sandramsc 👍 Another PR got merged that had light/dark theme changes also. Could you resolve the conflicts, thanks

pages/map.js
yoyoekta, line 1: I don't understand the change. Isn't it the same?

sandramsc: Lower case changed to upper case to maintain consistency.

yoyoekta: Good spot! Thanks for clarifying.

components/GitHubAccelerator.js
```diff
@@ -35,7 +35,7 @@ export default function GitHubAccelerator() {
             <div className="mt-6 flex flex-col justify-stretch space-y-3 sm:flex-row sm:space-x-4 sm:space-y-0">
               <Link
                 href="http://accelerator.github.com"
-                className="inline-flex justify-center rounded-md bg-white px-3 py-2 text-sm font-semibold text-primary-high dark:text-primary-high shadow-sm ring-1 ring-inset ring-primary-low-medium hover:bg-primary-low dark:hover:bg-slate-200 cursor-pointer"
+                className="inline-flex justify-center rounded-md bg-primary-low px-3 py-2 text-sm font-semibold text-primary-high dark:text-primary-high shadow-sm ring-1 ring-inset ring-primary-low/80 hover:bg-primary-low/95 dark:hover:bg-primary-low/95 cursor-pointer"
```
yoyoekta, line 38: Could you please tell what does this mean and how you come up with this? I want to know more about `/80` and `/95` thing.

`ring-primary-low-medium` to `ring-primary-low/80`
`bg-slate-200` to `bg-primary-low/95`

yoyoekta (approved):

components/GitHubAccelerator.js
sandramsc: Sure, I studied the previous code and understood the the logic behind the styling first, then implemented the change as per the tailwind.config.js requirements.

i.e 
**before**: `bg-white/5` , changing the numerical value in turn changes the shade of the color preceding it.

**after**: `bg-primary-low/85` (values starting ~ 5 to 100 will give different shades of the color associated with -primary-low (or any of the configured colors) in the tailwind.config.js file)

Hope that helps.

pages/newsletter.js
```diff
@@ -60,7 +60,7 @@ export default function Newsletter() {
                     </div>
                   </div>
                   <div className="flex flex-col items-start">
-                    <div className="rounded-md bg-primary-low dark:bg-primary-medium p-2 ring-1 ring-primary-medium">
+                    <div className="rounded-md bg-primary-low/25 dark:bg-primary-medium dark:ring-primary-medium-low p-2 ring-1 ring-primary-low/100">
```
ChinmayMhatre, line 63: I feel the border around both icons can be a better darker in light mode

components/Modal.js
ChinmayMhatre, line 1: image
The close icon is not visible in light mode and the section below should be a different gray

components/IconCard.js
ChinmayMhatre, line 1: image
Icon names not visible in light mode

components/IconCard.js
```diff
@@ -8,15 +8,15 @@ export default function IconCard({ iconName }) {
   return (
     <>
       <button
-        className="border-2 p-2 border-slate-100 w-24 h-24 flex flex-col items-center justify-around rounded hover:border-secondary-high hover:cursor-pointer active:border-green-600 focus:border-green-600"
+        className="border-2 p-2 border-primary-low/95 w-24 h-24 flex flex-col items-center justify-around rounded hover:border-secondary-high hover:cursor-pointer active:border-green-600 focus:border-green-600"
```
ChinmayMhatre, line 11: border can just be primary-low?

components/GitHubAccelerator.js
ChinmayMhatre: It changes the alpha of the colour which is the opacity not the actual shade. so `green-50/90` is `green-50` with 90 percent opacity

ChinmayMhatre: you can learn more here https://tailwindcss.com/docs/text-color . under the changing opacity section

components/GitHubAccelerator.js
ChinmayMhatre, line 1: image
The border around the buttons are not visible

ChinmayMhatre (changes requested):
Hi @sandramsc , Thank you for your effort and awesome work on the PR!🔥 I have left a few comments where the pages are breaking and a few other comments. Do check it out and let me know.

components/GitHubAccelerator.js
sandramsc: I didn't realize there was a difference between shade and opacity.

pages/newsletter.js
sandramsc: This opacity/shade was chosen to replicate the previous color `bg-white/5`. The goal of this PR is to replace the bg-[color] with its closest tailwind.config.js configuration. The comment above leans more towards changing the original design. I hope I understood your request correctly.
svg

ChinmayMhatre: You have changed the ring colour from `primary-medium` to `primary-low/100` which isn't required . you can see this on line 63 as marked by the comment. same is the case which `bg-primary-low`

sandramsc: - This change was made to fix the color adjustments from the first pull request within the same issue. Consider viewing the discussion here: https://github.com/EddieHubCommunity/LinkFree/pull/7782#discussion_r1241006010
- I've updated the opacity/shade again to colors that more closely replicate the original colors.
- My choices for the color choice is again purely to replicate the original (before) colors as close as possible to the colors available in the tailwind.config.js
- Here is how the colors look in my code editor, when comparing the **before** and **after** (hence the choice)
newletter_icon

sandramsc: - This change was made to fix the color adjustments from the first pull request within the same issue. Consider viewing the discussion here: https://github.com/EddieHubCommunity/LinkFree/pull/7782#discussion_r1241006010
- I've updated the opacity/shade again to colors that more closely replicate the original colors.
- My choices for the color choice is again purely to replicate the original (before) colors as close as possible to the colors available in the tailwind.config.js
- Here is how the colors look in my code editor, when comparing the **before** and **after** (hence the choice)
newsletter_icons

PR #7912, opened 2023-07-01
ci: Add vercel preview PR workflow

Pradumnasaraf:
## Changes proposed

- Create a Vercel Preview env with every PR that has a change in the JS file.

## Note to reviewers

The reason why I added branches-ignore is that if any preview branch is deployed to a custom domain, it will deploy twice - once with the custom domain and once with this action.

```yaml
    branches-ignore:    
      - revert-logger
```

<a href="https://gitpod.io/#https://github.com/EddieHubCommunity/LinkFree/pull/7912"></a>

### DIFF
--- a/.github/workflows/vercel-pr-preview.yml
+++ b/.github/workflows/vercel-pr-preview.yml
@@ -28,6 +28,7 @@ jobs:
           ls -l .vercel/output/functions
           vercel deploy --prebuilt --token=${{ secrets.VERCEL_TOKEN }} > domain.txt
           vercel alias --scope ${{ secrets.VERCEL_TEAM_ID }} --token ${{ secrets.VERCEL_TOKEN }}
+          echo "Deployed to $(cat domain.txt)"
   load-data:
     runs-on: ubuntu-latest
     needs: deploy
--- a/.github/workflows/vercel-pr-preview.yml
+++ b/.github/workflows/vercel-pr-preview.yml
@@ -8,6 +8,8 @@ on:
   pull_request:
     paths:
       - "**.js"
+    branches-ignore:    
+      - revert-logger
 
 jobs:
   deploy:
--- a/.github/workflows/vercel-pr-preview.yml
+++ b/.github/workflows/vercel-pr-preview.yml
@@ -6,10 +6,10 @@ env:
 on:
   workflow_dispatch:
   pull_request:
-    paths:
-      - "**.js"
     branches-ignore:    
       - revert-logger
+    paths:
+      - "**.js"
 
 jobs:
   deploy:
--- a/.github/workflows/vercel-pr-preview.yml
+++ b/.github/workflows/vercel-pr-preview.yml
@@ -1,0 +1,35 @@
+name: Deploy PR to Preview
+env:
+  VERCEL_ORG_ID: ${{ secrets.VERCEL_ORG_ID }}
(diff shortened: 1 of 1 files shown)


eddiejaoude (changes requested):
Does this Action now replace the existing preview Action? https://github.com/EddieHubCommunity/LinkFree/blob/main/.github/workflows/vercel-preview.yml

Pradumnasaraf:
> Does this Action now replace the existing preview Action? https://github.com/EddieHubCommunity/LinkFree/blob/main/.github/workflows/vercel-preview.yml

No, we will keep that as well so that we can deploy certain branch to set up a custom domain.

.github/workflows/vercel-pr-preview.yml
```diff
@@ -0,0 +1,38 @@
...
+      - name: pull Vercel environment information
+        run: vercel pull --yes --environment=preview --token=${{ secrets.VERCEL_TOKEN }}
+      - name: build project artifacts
+        run: vercel build --token=${{ secrets.VERCEL_TOKEN }}
+      - name: deploy preview + assign beta domain
+        run: |
+          du --inodes -d 5 .vercel/output
+          ls -l .vercel/output/functions
+          vercel deploy --prebuilt --token=${{ secrets.VERCEL_TOKEN }} > domain.txt
+          vercel alias --scope ${{ secrets.VERCEL_TEAM_ID }} --token ${{ secrets.VERCEL_TOKEN }}
```
dan-mba, line 30: What is this supposed to do?
It doesn't have one of the commands after alias like in the docs.
https://vercel.com/docs/cli/alias

.github/workflows/vercel-pr-preview.yml
```diff
@@ -0,0 +1,38 @@
...
+          echo "Deployed to $(cat domain.txt)"
+  load-data:
+    runs-on: ubuntu-latest
+    needs: deploy
+    steps:
+      - name: load json files
+        if: steps.changes.outputs.data == 'true'
+        run: curl -f https://linkfree-preview.vercel.app/api/system/reload?secret=${{ secrets.LINKFREE_API_SECRET_PREVIEW }}
```
dan-mba, line 38: Shouldn't this point to the URL from the `domain.txt` file?
I think the if statement can be removed also.

dan-mba:
Would it make sense to add running out playwright tests on the preview deployment?
Or should we add that in a later PR?

.github/workflows/vercel-pr-preview.yml
Pradumnasaraf: Oh, yes. I was previous doing a domain assigning linkfree-commitid.vercel.app then plan to keep it simple.

I have removed it. Thank you.

.github/workflows/vercel-pr-preview.yml
Pradumnasaraf: Good catch. Fixed. Thank you.

Pradumnasaraf:
> Would it make sense to add running out playwright tests on the preview deployment? Or should we add that in a later PR?

Ya, it will be good. We can do this in another PR. Also, thanks for reviw

dan-mba (approved):
Looks good to me.

eddiejaoude (approved):
Looks good 👍

PR #7932, opened 2023-07-02
[fix] for "GitHub Accelerator" and "GitHub Sponsors" buttons alignment

### LINKED ISSUE #7914: [BUG] "GitHub Accelerator" and "GitHub Sponsors" buttons are not horizontally aligned - mobile devices
### Description

The "GitHub Accelerator" and "GitHub Sponsors" buttons are not ("perfectly") horizontally aligned on mobile devices.

### Screenshots

gacc

### Additional information

_No response_

sandramsc:
## Fixes Issue
Closes #7914

## Screenshots
**Before:**
before

**After:**
after

<a href="https://gitpod.io/#https://github.com/EddieHubCommunity/LinkFree/pull/7932"></a>


components/GitHubAccelerator.js
```diff
@@ -19,7 +19,7 @@ export default function GitHubAccelerator() {
         <div className="-mt-12 sm:-mt-16 sm:flex sm:items-end sm:space-x-5">
           <div className="flex">
             <Image
-              className="h-24 w-24 rounded-full ring-4 ring-white sm:h-32 sm:w-32"
+              className="h-24 w-24 rounded-full ring-4 ring-primary-low sm:h-32 sm:w-32"
```
eddiejaoude, line 22: Good spot 👍 Do we need a `dark` version?

components/GitHubAccelerator.js
```diff
@@ -35,25 +35,25 @@ export default function GitHubAccelerator() {
             <div className="mt-6 flex flex-col justify-stretch space-y-3 sm:flex-row sm:space-x-4 sm:space-y-0">
               <Link
                 href="http://accelerator.github.com"
-                className="inline-flex justify-center rounded-md bg-white px-3 py-2 text-sm font-semibold text-primary-high dark:text-primary-high shadow-sm ring-1 ring-inset ring-primary-low-medium hover:bg-primary-low dark:hover:bg-slate-200 cursor-pointer"
+                className="inline-flex ml-1.7 justify-center rounded-md bg-white px-3 py-2 text-sm font-semibold text-primary-high dark:text-primary-high shadow-sm ring-1 ring-inset ring-primary-low-medium hover:bg-primary-low dark:hover:bg-slate-200 cursor-pointer"
```
eddiejaoude, line 38: We should not try to sign items with margins, because as the length of the content changes so will the alignment requirements

It is best use alignment, at the moment it is `justify-center` but for smaller screens I think from your screenshot, left align would solve this

eddiejaoude (changes requested):
Thank you 👍 

I left inline comments to use alignment instead of margins for positioning

components/GitHubAccelerator.js
sandramsc: Hi Eddie, these changes (and others) were made in this PR, but it seems it wasn't merged so the changes were not applied to the repo.

Consider viewing the related repo [here](https://github.com/EddieHubCommunity/LinkFree/pull/7791)

components/GitHubAccelerator.js
sandramsc: Thank you for clarifying this, I've removed the margin alignment (ml-1.7 ).

components/GitHubAccelerator.js
eddiejaoude: Ok, please can you remove those changes that are not part of this PR

sandramsc: Requested change has been implemented.

eddiejaoude:
Let us know if you need help with this

sandramsc:
> Let us know if you need help with this

Hi Eddie,

When I add **sm** to the margin left of the GitHub Accelerator button it does not take the expected effect i.e rendering the assigned value on small devices. Using **justify-center** also does not produce the desired out come, which is for the content of both buttons to align vertically starting with the logos as in screenshot. 

Since GitHub Accelerator button content is longer in length than the GitHub Sponsors button, the best solution I used (and know of at this point) was to add an **lg** to the original margin left value and add an additional margin left (without lg or sm assignment) that seems to only effect the GitHub Accelerator button on the small screen without the sm value (and produces the desired outcome).

Better approaches/assistance/suggestions are welcome.

eddiejaoude:
Ah yes, if we align left it looks worse, but we can't use margins, because if we change the text then it will be even more broken. I think the best for now is to close this PR and Issue, as we start the whole new process of the rebrand from next week. Thank you for your efforts with this but I think your energy would be better used on other improvements to the project.

Screenshot 2023-07-11 at 22 15 21

PR #8056, opened 2023-07-08
fix: Event Card price icon bug

### LINKED ISSUE #7854: [BUG] Ticket price not working correctly
### Description

The event block looks different on the list and edit page. It stops adding a dollar sign on the preview on the right side. It should also it needs to show up correctly on the events page.
 

### Screenshots

Screenshot_2023-06-28_at_3 47 19_PMScreenshot_2023-06-28_at_3 36 01_PM
Screenshot_2023-06-28_at_3 37 02_PM

### Additional information

_No response_

itsdevdeepak:
## Fixes Issue

Closes #7854 

## Changes proposed
- updated `getEvents()` method to include price
- updated manage event page to handle edge cases

## Note to reviewers

<a href="https://gitpod.io/#https://github.com/EddieHubCommunity/LinkFree/pull/8056"></a>


pages/account/manage/event/[[...data]].js
```diff
@@ -79,7 +79,7 @@ export default function ManageEvent({ BASE_URL, event }) {
       url,
       date: { start: startDate, end: endDate },
       isVirtual,
-      price: { startingFrom: price },
+      price: { startingFrom: price || 0 },
```
Comment (amandamartin-dev) [inline pages/account/manage/event/[[...data]].js:82]: I would disagree that the price should be set to zero in the case of no price.  What if the price was unknown or simply not entered by the event submitter? I would suggest that we do not show price if it is unknown as this could be misleading or incorrect. wdyt?

Comment (eddiejaoude) [reply pages/account/manage/event/[[...data]].js:82]: Oh good point - we can then display `unknown` if there is no value

Comment (itsdevdeepak) [reply pages/account/manage/event/[[...data]].js]: sorry for the late commit but i don't why mongodb not connection with docker compose, in the end i push the commit with git pod

eddiejaoude (changes requested):
Looks good 👍  but there still seems to be a bug on the main events page:

Updated event with price which works fine...

Screenshot 2023-07-12 at 07 50 42

But on the main event page it is not reflected correctly...

Screenshot 2023-07-12 at 07 51 10

itsdevdeepak:
Closing it due to conflict

eddiejaoude:
Ok, but it is better to resolve the conflicts and keep it in the same PR, because the history and collaboration is lost in the new PR

itsdevdeepak:
> Ok, but it is better to resolve the conflicts and keep it in the same PR, because the history and collaboration is lost in the new PR

I will keep that in mind for next time.

itsdevdeepak:
Reopening pr

itsdevdeepak:
> Looks good +1 but there still seems to be a bug on the main events page:
> 
> Updated event with price which works fine...
> Screenshot 2023-07-12 at 07 50 42
> 
> But on the main event page it is not reflected correctly...
> Screenshot 2023-07-12 at 07 51 10

This bug is related to the current workflow of event creation, right now if you create a event with a same event name that is already exist, It creates the new event with different id and getEvents method groups them but all the other properties like price, description etc are getting its value from event that is declared $first.

I could make it something like $last or fist non null but it will make it inconsistent.
what do you think @eddiejaoude ?

eddiejaoude:
> I could make it something like $last or fist non null but it will make it inconsistent.
what do you think @eddiejaoude ?

I think leave that for now, as long as a unique event works, we can look into improving the grouping in another issue/pr

adityaraute (approved):

eddiejaoude (approved):
Looks good 👍

PR #8087, opened 2023-07-09
[fix]: comma misbehave in profile form section

### LINKED ISSUE #8040: [BUG] comma's misbehave in the form, editor tags section
### Description

Comma at the end of the tags should be ignored 
image

Also, with a comma at the end make one of the central comma's a double comma ",," will give you an empty tag in the middle.  When you then delete one of the extra comma's the empty tag never goes away.
image

### Screenshots

_No response_

### Additional information

_No response_

Ashish-simpleCoder:
## Fixes Issue

Closes #8040 

## Changes proposed
- Empty blocks in the tag list won't appear anymore.
- Updating the tags with the response of the profile-update api.
- Updated the key prop's value to the ```index```, if same-named tags are entered multiple times and also for removing them when user removes them from the tag-input field section. Rendered ```Tag``` Components will not unmount if new tags are added.

- Backend API Changes -> Added filter for removing all of the falsy value and string values which contains only whitespaces to prevent whitespaced link generation for the tags when user clicks on them.

Example
if input is ```" javascript,          nextjs        ,,,,,solidjs, vuejs,"```, then they will be result into the ```"javascript, nextjs,solidjs,vuejs"``` while saving in db.

## Screenshots

This is the screenshot of the tags before saving them.
before_save_tags

This is the screenshot after saving the profile changes.
after_save_tags

## Note to reviewers

<a href="https://gitpod.io/#https://github.com/EddieHubCommunity/LinkFree/pull/8087"></a>

### DIFF
--- a/components/user/UserProfile.js
+++ b/components/user/UserProfile.js
@@ -76,15 +76,19 @@ function UserProfile({ BASE_URL, data }) {
       {!qrShow && (
         <div className="flex flex-wrap justify-center">
           {data.tags?.length > 0 &&
-            data.tags.map((tag) => (
-              <Tag
-                name={tag}
-                key={tag.toLowerCase()}
-                onClick={() =>
-                  router.push(`/search?keyword=${tag.toLowerCase()}`)
-                }
-              />
-            ))}
+            data.tags.map((tag, index) => {
+              const trimmed_tag = tag.trim();
+              if (!trimmed_tag) return null;
+              return (
+                <Tag
+                  name={trimmed_tag}
+                  key={index}
+                  onClick={() =>
+                    router.push(`/search?keyword=${trimmed_tag.toLowerCase()}`)
+                  }
+                />
+              );
+            })}
         </div>
       )}
 
--- a/components/user/UserProfile.js
+++ b/components/user/UserProfile.js
@@ -78,7 +78,9 @@ function UserProfile({ BASE_URL, data }) {
           {data.tags?.length > 0 &&
             data.tags.map((tag, index) => {
               const trimmedTag = tag.trim();
-              if (!trimmedTag) return null;
+              if (!trimmedTag) {
+                return null;
+              }
               return (
                 <Tag
                   name={trimmedTag}
--- a/components/user/UserProfile.js
+++ b/components/user/UserProfile.js
@@ -77,14 +77,14 @@ function UserProfile({ BASE_URL, data }) {
         <div className="flex flex-wrap justify-center">
           {data.tags?.length > 0 &&
             data.tags.map((tag, index) => {
-              const trimmed_tag = tag.trim();
-              if (!trimmed_tag) return null;
+              const trimmedTag = tag.trim();
+              if (!trimmedTag) return null;
               return (
                 <Tag
-                  name={trimmed_tag}
+                  name={trimmedTag}
                   key={index}
                   onClick={() =>
-                    router.push(`/search?keyword=${trimmed_tag.toLowerCase()}`)
+                    router.push(`/search?keyword=${trimmedTag.toLowerCase()}`)
                   }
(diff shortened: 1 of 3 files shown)


components/user/UserProfile.js
```diff
@@ -76,15 +76,19 @@ function UserProfile({ BASE_URL, data }) {
...
-              <Tag
-                name={tag}
-                key={tag.toLowerCase()}
-                onClick={() =>
-                  router.push(`/search?keyword=${tag.toLowerCase()}`)
-                }
-              />
-            ))}
+            data.tags.map((tag, index) => {
+              const trimmed_tag = tag.trim();
```
eddiejaoude, line 80: the naming does not follow the project standards, it should be `trimmedTag`

components/user/UserProfile.js
```diff
@@ -76,15 +76,19 @@ function UserProfile({ BASE_URL, data }) {
...
+              if (!trimmed_tag) return null;
```
eddiejaoude, line 81: this style does not follow the project standard, we use `{}` for conditions

components/user/UserProfile.js
```diff
@@ -76,15 +76,19 @@ function UserProfile({ BASE_URL, data }) {
...
-            data.tags.map((tag) => (
-              <Tag
-                name={tag}
-                key={tag.toLowerCase()}
-                onClick={() =>
-                  router.push(`/search?keyword=${tag.toLowerCase()}`)
-                }
-              />
-            ))}
+            data.tags.map((tag, index) => {
```
eddiejaoude, line 79: Maybe a `flatMap` would be better here, so items can be removed?

pages/account/manage/profile.js
```diff
@@ -93,6 +93,7 @@ export default function Profile({ BASE_URL, profile, fileExists }) {
         ).join(", ")}`,
       });
     }
+    setTags(update.tags);
```
eddiejaoude, line 96: good spot 👍

pages/api/account/manage/profile.js
```diff
@@ -57,7 +57,9 @@ export async function updateProfileApi(username, data) {
     layout: data.layout,
     name: data.name,
     bio: data.bio,
-    tags: data.tags.filter((tag) => tag !== ""),
+    tags: data.tags
+      .filter((tag) => Boolean(tag.trim()))
```
eddiejaoude, line 61: Not sure what this is for?

eddiejaoude (changes requested):
Thank you 👍 I left some inline comments

components/user/UserProfile.js
Ashish-simpleCoder: I am sorry. I will change this.

components/user/UserProfile.js
Ashish-simpleCoder: There is no need for flatting the array , we only need to remove falsy values.
We can also filter the items before mapping , but that will result into twice iteration over the tags list.
That's why I have choosen that approach.

pages/api/account/manage/profile.js
Ashish-simpleCoder: Are you talking about Array.filter(item => Boolean(item))?
This is filter technique, which can filter out any falsy values from an array. Values like null, undefined, false , "" etc which are falsy in Javascript, they will be removed from array.

Ashish-simpleCoder: Prevoius code was not handling cases for strings like ```"     "```, which contains whitespace.

eddiejaoude: ok thank you! Please don't resolve in the future, otherwise I might miss your replies

eddiejaoude (approved):
Looks good 👍  thank you!

PR #7522, opened 2023-06-11
Community events to map

dhruvilmehta:
## Fixes Issue

fixes #4789 

## Changes proposed
First of all there are many changes I am proposing to successfully solve this issue.
1. Issue #4132 with  Pull request #4438 was closed by adding 4 location parameters(road, city, state, country) but to locate the event into Map we need 2 more parameters which is Latitude and Longitude so I have added that 2 parameters in the events section. 
  For testing purposes I have added these 2 parameters in this event "data\TRohit20\events\2023-06-13-DevRelCon-London.json"
2. Also, the pull request I mentioned was incomplete because though he had added the location parameters in the event, he had not made that changes into the mongoose model and hence the location was not going into database.
3. I have made the necessary changes to store the location parameter into the mongoose model.
4. I have fetched the events from "pages/map.js" and further passed that events to "components/Map.js"
5. I have also added an event icon in the "public" folder to show in the map that there is an event. (see the screenshot added)
6. Done necessary changes in "events" documentation.

## Check List (Check all the applicable boxes) 

- [x] My code follows the code style of this project.
- [x] My change requires changes to the documentation.
- [x] I have updated the documentation accordingly.
- [ ] All new and existing tests passed.
- [x] This PR does not contain plagiarized content.
- [x] The title of my pull request is a short description of the requested changes.

## Screenshots
Screenshot (42)
This is how the location parameters look like (i have added "lat" and "lon" into this to locate this in map)
-------------------------------------------------------
Screenshot (46)

Screenshot (45)
As soon as we click into the event shown on map we can see the details of the event.
-------------------------------------------------------

## Note to reviewers

<a href="https://gitpod.io/#https://github.com/EddieHubCommunity/LinkFree/pull/7522"></a>

PR #8093, opened 2023-07-09
feat: community events to map (#7522)

eddiejaoude:
Finalising PR https://github.com/EddieHubCommunity/LinkFree/pull/7522

Items I am looking at: (I will keep updating this list as I work through the PR)

- [x] remove image for event and use icon, fix in commit  https://github.com/EddieHubCommunity/LinkFree/pull/7522/commits/c676b9f3ff124d0f73f4d20d349ce65fdb695ba9
   - [ ] ~~does the event need a logo?~~ future idea
   - [ ] ~~if virtual use icon~~ future idea
   - [ ] ~~if in person use icon~~ future idea
- [ ] map to have toggle for users/events (do events need tags also?)
- [ ] refactor query to be more readable
- [ ] refactor the way location is done to remove duplicate with user location


amandamartin-dev:
Adding a note here for consideration (now or future) Leaflet has an ootb functionality to create layers for data so that folks can quickly toggle the view based on what they want to see.  Because we are using an external to leaflet filter, we may or may not want to use this functionality.  Here is a link to check it out - happy to open a future issue if we like this direction https://react-leaflet.js.org/docs/example-layers-control/

PR #8145, opened 2023-07-11
fix: protected pages tests

eddiejaoude:
## Fixes Issue

- [x] test guests cannot access logged in pages
- [x] logged in user cannot access admin page 
- [ ] admin user can access admin page


tests/setup/auth.js
```diff
@@ -4,18 +4,23 @@ import { encode } from "next-auth/jwt";
...
   const date = new Date();
   const sessionToken = await encode({
     token: {
-      name: "Automate Test",
-      email: "testemail@test.com",
       image: "https://github.com/eddiejaoude.png",
       access_token: "ggg_zZl1pWIvKkf3UDynZ09zLvuyZsm1yC0YoRPt",
-      username: "eddiejaoude",
-      id: "22222222",
+      ...user,
```
dan-mba, line 23: Switching the username from `eddiejaoude` to `_test-profile-user-1` will effect all the manage tests. This test user profile need to be updated to make use of events, milestones, testimonials, etc. so the tests will check all these features. That is why we tended to test with your profile, because you take advantage of every feature.

eddiejaoude: Good spot, thank you 👍 I think user 6 has all the features enabled, I will try with that

PR #8191, opened 2023-07-14
Feat: Search Persistance

### LINKED ISSUE #8109: [FEATURE] Search persistence
### Description

When exploring the results of a search I have to re-enter my search terms every time as the search does not persist.  This is cumbersome if I want to click on many profiles to explore within a single search.

update per conversation in thread below:  Current approach to test is persisting the search with url query params for an effective/simple solution.

~~Offering two potential solutions:~~
~~1. Persist a search term when clicking "back" on the browser~~
~~2. Have profiles open in a new tab from the search page so that it does not disrupt the current search parameters~~

Video of issue:

https://github.com/EddieHubCommunity/LinkFree/assets/97615019/d344cf59-a8ba-4c9b-ab02-1a825cd5a5b0

### Screenshots

_No response_

### Additional information

_No response_

yashug:
## Fixes Issue

Closes #8109 

## Changes proposed

- replaces URL whenever a new fetch request goes
- uses window API to replace URL instead of next router to avoid rerendering of the page
- removes unwanted console log

## Screenshots

Screenshot 2023-07-14 at 8 00 56 PM

## Note to reviewers

<a href="https://gitpod.io/#https://github.com/EddieHubCommunity/LinkFree/pull/8191"></a>

### DIFF
--- a/pages/search.js
+++ b/pages/search.js
@@ -49,14 +49,12 @@ export default function Search({ data: { tags, randUsers }, BASE_URL }) {
   const { username, keyword } = router.query;
   const [notFound, setNotFound] = useState();
   const [users, setUsers] = useState(randUsers);
-  const [inputValue, setInputValue] = useState(username || keyword || "");
+  const [inputValue, setInputValue] = useState("");
   const [currentPage, setCurrentPage] = useState(1);
 
   useEffect(() => {
-    if (username) {
-      setNotFound(`${username} not found`);
-    }
-  }, [username]);
+    setInputValue(username || keyword || "");
+  }, [username, keyword]);
 
   useEffect(() => {
     if (!inputValue) {
@@ -95,6 +93,14 @@ export default function Search({ data: { tags, randUsers }, BASE_URL }) {
 
     const timer = setTimeout(() => {
       fetchUsers(inputValue);
+      const newUrl = `/search?keyword=${inputValue}`;
+
+      // replace the url with the new url without reloading the page
+      window.history.replaceState(
+        { ...window.history.state, as: newUrl, url: newUrl },
+        "",
+        newUrl
+      );
     }, 500);
 
     return () => clearTimeout(timer);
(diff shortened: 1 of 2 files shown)


pages/search.js
```diff
@@ -95,6 +93,14 @@ export default function Search({ data: { tags, randUsers }, BASE_URL }) {
 
     const timer = setTimeout(() => {
       fetchUsers(inputValue);
+      const newUrl = `/search?keyword=${inputValue}`;
+
+      // replace the url with the new url without reloading the page
+      window.history.replaceState(
```
eddiejaoude, line 99: I think react has builtin functionality for this

eddiejaoude (changes requested):
Thank you for your contribution, I have left an inline comment

pages/search.js
yashug: Thank you for your response. I wasn't aware of any built-in functionality in React for updating the URL without reloading the page. I tried using Next.js' router.replace, but it reloads the page, causing redundant API calls and useEffect re-renders. Could you please provide more information or guidance on the React built-in functionality you mentioned? Any specific resources would be greatly appreciated.

yashug: @eddiejaoude can we please have a update on this

yashug: router push will reload the page and also add a new entry to the browser history stack. if we don't need to bother about extra reload of the page, router `replace` would be the best choice as it won't add a new entry into the history stack. please let me know what I can follow here.

yashug:
@Ashish-simpleCoder as I see, you have added a comment on the issue before, using router` replace`. What's your thought on my approach here? my main goal is to not rerender the page again on route change, thus minimize unnecessary API call and rerenders

Ashish-simpleCoder:
What do you mean by page rerendering?
Is the whole page is reloading when using router.replace from nextjs?
Correct?

pages/search.js
eddiejaoude: Yes I think it is better to go onto the history stack.

> Shallow routing allows you to change the URL without running data fetching methods again. The URL will get updated to /?counter=10 and the page won't get replaced, only the state of the route is changed.

You can use `push` without a page reload using `shallow` option, you can read the official docs here https://nextjs.org/docs/pages/building-your-application/routing/linking-and-navigating#shallow-routing

yashug: already tried using push with shallow option

yashug:
> @yashug
> 
> What do you mean by page rerendering? Is the whole page is reloading when using router.replace from nextjs? Correct?

yes

pages/search.js
eddiejaoude: We could move it out of the `use effect` into a function that is triggered by the input field directly?

Ashish-simpleCoder:
> > @yashug
> > What do you mean by page rerendering? Is the whole page is reloading when using router.replace from nextjs? Correct?
> 
> yes

Can you share just the small video and code that you have written for managing the url update using NextJS Router?
NextRouter should not be reloading the tab.

SaraJaoude:
Due to lack of activity I am closing this PR.

PR #8256, opened 2023-07-18
Simple tags for user cards on search page

### LINKED ISSUE #8217: [FEATURE] Add tags to user cards on search page
### Description

From a conversation in Discord...

Searching by tag often produces confusion in the results as the tags are not visible in the cards.  This creates the perception that the search may be functioning incorrectly.

Solution:
Add tags to the profile cards

### Screenshots

_No response_

### Additional information

_No response_

amandamartin-dev:
## Fixes Issue #8217 

 Closes #8217 

## Changes proposed

added a new tag component stripped for the user cards
added a horizontal version of the user cards with tags

follow up issue to be created for this after merge to highlight tags based on search terms

- [] My change requires changes to the documentation.

## Note to reviewers

<a href="https://gitpod.io/#https://github.com/EddieHubCommunity/LinkFree/pull/8256"></a>


amandamartin-dev:
@eddiejaoude or anyone passing through, still playing with this but wanted to show the first iteration.  As there are numerous ways folks use tags, some cards work well and otehrs do not

One idea to address this is to limit the amount of tags shown on the card, however, in some cases this may confuse the users during search which is what the point of this issue is

Let me know your thoughts/ideas so far

Screenshot
draft-cards

amandamartin-dev:
I also would love to hear thoughts about changing the color to the border.  I kept it the same as the Tag component for continuity, but with a smaller visual area, some color to break things up could look better. Let me know your thoughts on overall style as well

accodes21:
I agree with the idea of keeping the tags to a limited number as it would create uniformity, imo users should see to it that which tags come on their profile **but** it might get tedious to implement so ig let's keep the first 3 or 4 tags as the user themselves entered them in that priority order.
This is just a suggestion, others might have a different view on it.

sandramsc:
> I agree with the idea of keeping the tags to a limited number as it would create uniformity, imo users should see to it that which tags come on their profile **but** it might get tedious to implement so ig let's keep the first 3 or 4 tags as the user themselves entered them in that priority order. This is just a suggestion, others might have a different view on it.

- I agree and to add to this it makes the overall content within the frame look cleaner.
- Effort might need to be taken however in informing the profile creators at first to put their most favoured links first if this approach is taken, but at the same time it could be phrased as '..add strongest or most proficient skills first'.
- I see this as being the better approach unless there is a standard that can be used in truncating/abbreviating the tag names (resulting in short tags) so individuals have the option to add more than a specified number, however even with this approach ... too many short tags could be added, resulting in the issue not being resolved.

components/user/UserCard.js
```diff
@@ -13,21 +14,36 @@ export default function UserCard({ profile }) {
...
-          height={80}
+          width={75}
+          height={75}
           className="rounded-full"
           fallback={profile.name}
         />
       </div>
       <div>
-        <h3 className="text-xl justify-center text-center mb-2 text-tertiary-medium font-bold">
+        <h3 className="text-lg justify-center text-center mb-2 text-tertiary-medium font-bold">
```
dan-mba, line 24: Because of the decrease in font size, the color need to be changed to increase contrast on a light background.

amandamartin-dev: This is a very good point.  WDYT about the overall concept so far though?  Trying to decide if I should stick with this direction or perhaps a different iteration/style of card

dan-mba: I'm not sure if there is a great way to show all tags and the cards look balanced.
I dealt with this problem on my portfolio site & never found a good solution.

If we are only showing tags that match a search term, then I think this will look OK, maybe with a accent color for the border.

amandamartin-dev: Yeah that's where my mind is going too.  Show tags based on the search terms (up to X amount so they fit) since this issue is about making the search make sense from a user perspective.  Thanks!

ChinmayMhatre:
@amandamartin-dev what do you think about a 'search by tag'/'search by user' toggle at the top?

eddiejaoude:
Sorry for missing the notification Amanda! This is a challenge with the same, could I throw a curve ball and suggest using a different card layout? Maybe making the profile pic smaller and to the size, this is from one of the tailwind paid examples (let me know if you would like me to share the code)

Screenshot 2023-07-23 at 09 10 35


https://tailwindui.com/components/application-ui/lists/grid-lists#component-2beafc928684743ff886c0b164edb126

amandamartin-dev:
> @amandamartin-dev what do you think about a 'search by tag'/'search by user' toggle at the top?

Hey Chimnay! We already have the tag search, the issue is you don't see tags in the profile cards so it's confusing. Unless I'm misunderstanding your suggestion

amandamartin-dev:
> Sorry for missing the notification Amanda! This is a challenge with the same, could I throw a curve ball and suggest using a different card layout? Maybe making the profile pic smaller and to the size, this is from one of the tailwind paid examples (let me know if you would like me to share the code)
> 
> Screenshot 2023-07-23 at 09 10 35
> https://tailwindui.com/components/application-ui/lists/grid-lists#component-2beafc928684743ff886c0b164edb126

I think it can't hurt to mock up a few options in this case. This isn't critical so I think okay if we take some time on the right UI. I'll check it out! I think I still have the log in for the components

eddiejaoude:
Thanks Amanda, have a look at the tailwind components and let me know which ones you would like the code for

amandamartin-dev:
@eddiejaoude tested a horizontal mock.  I think the vertical is still the best option here because the horizontal looks very empty when the profile has no tags.

Thinking of going with Dan's suggestion to show only tags that match search terms as part of the output (up to 3 or whatever fits well) and keeping the vertical card.

Can also abandon this route altogether if it's just getting to be too much in the cards. Lmk what you thinK!

horizontal-test

eddiejaoude (changes requested):
Looks good 👍

Some tags get cut off, do we want to limit the amount of tags shown?

Screenshot 2023-07-25 at 08 16 15

amandamartin-dev:
@eddiejaoude As I looked into this, it was impossible to decide where to split the tags as folks choose varying word length for tags.  As the data is not all the same, rather than split at an arbitrary amount of tags I chose to solve this by setting a max height on the div containing the tags and set the overflow to hidden.  This allowed for the most tags possible to show based on the tag size.

If you prefer me to also split the data out, I can but I will still leave this code in place as the number of tags could still be too long for the container if the character length per tag is long.

Take a look and let me know! Ready for review

components/TagSimple.js
```diff
@@ -0,0 +1,8 @@
+
+export default function TagSimple({ name }) {
+  return (
+  <div className="text-sm px-1 mx-px mb-2 font-mono border rounded-md line-clamp-1" >
+     {name}
+  </div>
+  );
+}
```
eddiejaoude, line 8: Now we have 2 tag components, should we use a `tag` folder and have them both in there?

eddiejaoude:
Looks great 👍 , just 1 small comment

components/user/UserHorizontal.js
```diff
@@ -0,0 +1,54 @@
+import ReactMarkdown from "react-markdown";
+
+import Link from "@components/Link";
+import FallbackImage from "@components/FallbackImage";
+import TagSimple from "@components/TagSimple";
+
+export default function UserHorizontal({ profile }) {
```
eddiejaoude, line 7: I don't think `UserCard` is used anymore, we should probably remove it

eddiejaoude (approved):
Looks great! 👍  I will merge to a temporary branch and make 2 little tweaks, thank you 🔥

PR #8283, opened 2023-07-20
fix: improve error handling  with HTTP utility functions

### LINKED ISSUE #8112: [BUG] needs proper error handling in manage form section
### Description

We have lots of AJAX call in `pages/account/manage` section and we are calling `fetch()` function. The problem is here that by default the `fetch` function doesn't reject the promise if we get back an error status code, in that case, we should reject the promise and throw an actual error and handle the rejected promise.

The downside of not handling a rejected promise in `fetch` call is that if the request fails for any reason, such as a network error or server-side error, the code does not handle the error gracefully. As a result, the application may not provide appropriate feedback or error handling to the user.

Without proper error handling, it becomes challenging to identify the cause of the issue. Debugging the code and fixing the problem becomes more complex, as there is no specific error message or indication of what went wrong.

To address this issue, it is important to handle rejected promises appropriately. This can involve displaying error messages to the user.

### Screenshots

_No response_

### Additional information

want to fix this issue

badrivlog:
## Fixes Issue
Closes #8112 

## What?
added a new util file `apiRequests.js`, containing functions for sending API requests.
Implemented `sendRequest` function that handles various types of API requests, handles rejected promise and also converts the `response` to `JSON`

## Why? 
abstract `api` request functionality into one function that we can use all over our project.
These changes improved error handling

## How?
replace `fetch` with `sendRequest` in manage profile sections

## Testing? 
**Step to test**
* Pull down this branch
* visit `localhost:3000/account/manage/profile` (link, milestone, event, testimonials) 
* send an invalid api request by `sendRequest` function
* see the request actually rejected and displayed the error to UI

## Anything else?
Let's consider implementing it other areas of the project. I will happy to doing that work💪

## Note to reviewers

<a href="https://gitpod.io/#https://github.com/EddieHubCommunity/LinkFree/pull/8283"></a>


badrivlog:
Hey, could I have some update regarding this PR?

eddiejaoude:
> Hey, could I have some update regarding this PR?

Sorry for the delay, I have shared with the maintainers and I am waiting to hear their thoughts on it

pages/account/manage/event/[[...data]].js
```diff
@@ -89,51 +90,51 @@ export default function ManageEvent({ BASE_URL, event }) {
...
-        "Content-Type": "application/json",
-      },
-      body: JSON.stringify(putEvent),
-    });
-    const update = await res.json();
+    try {
+      await sendRequest({
+        url: apiUrl,
+        method: "PUT",
+        headers: {
```
Comment (azhechkova) [inline pages/account/manage/event/[[...data]].js:97]: I think you can move header `Content-type` to sendRequest util as default header

pages/account/manage/event/[[...data]].js
```diff
@@ -89,51 +90,51 @@ export default function ManageEvent({ BASE_URL, event }) {
...
-    Router.push(`${BASE_URL}/account/manage/events?success=true`);
   };
 
   const deleteItem = async () => {
-    const res = await fetch(
-      `${BASE_URL}/api/account/manage/event/${event._id}`,
-      {
+    try {
+      await sendRequest({
+        url: `${BASE_URL}/api/account/manage/event/${event._id}`,
```
Comment (azhechkova) [inline pages/account/manage/event/[[...data]].js:123]: I'm not sure why BASE_URL comes from server side, as it's in the config. Maybe, it can be added to api client from the config and not passed through many components

pages/account/manage/profile.js
```diff
@@ -74,33 +75,39 @@ export default function Profile({ BASE_URL, profile, fileExists }) {
...
+      setTags(update.tags);
+      setShowNotification({
+        show: true,
+        type: "success",
+        message: "Profile updated",
+        additionalMessage: "Your profile has been updated successfully",
+      });
+    } catch (err) {
+      const errMessage =
+        typeof err === "string"
```
azhechkova, line 98: Duplicate logic as in `milestone/[[...data]]` file, I think it can be moved to some function (f.e. formatError)

pages/account/manage/testimonials.js
```diff
@@ -47,41 +48,75 @@ export async function getServerSideProps(context) {
...
+        url: `${BASE_URL}/api/account/manage/testimonials`,
+        method: "PUT",
+        headers: {
+          "Content-Type": "application/json",
+        },
+        body: testimonialList,
+      });
+      setTestimonialListPrevious(updatedTestimonials);
+      setReorder(false);
+      setShowNotification({
```
azhechkova, line 105: Duplicate logic as on 75-83 lines (the same in catch block). I suggest adding separate functions f.e. `
const updateSuccess = () => {
setShowNotification({
        show: true,
        type: "success",
        message: "Testimonials updated",
        additionalMessage:
          "Your profile information has been saved successfully.",
      });
}
`

pages/account/manage/event/[[...data]].js
Comment (badrivlog) [reply pages/account/manage/event/[[...data]].js:97]: Good idea 👍

pages/account/manage/event/[[...data]].js
Comment (badrivlog) [reply pages/account/manage/event/[[...data]].js:123]: You're right 👍

pages/account/manage/profile.js
badrivlog: Great idea👍

pages/account/manage/testimonials.js
badrivlog: OMG that's great idea 👍 
Thank you for review my PR.

badrivlog:
@azhechkova,

I'd greatly appreciate it if you could take another look at the pull request and let me know if the changes align with your expectations. Your expertise is truly valued in ensuring the quality of the code.


Thank you for your time and consideration!

azhechkova:
@badrivlog sorry for the delay, PR looks good!

azhechkova (approved):

badrivlog:
> @badrivlog sorry for the delay, PR looks good! 

No problem at all! Thanks for taking the time to review.

services/utils/apiRequests.js
```diff
@@ -0,0 +1,32 @@
+import { clientEnv } from "@config/schemas/clientSchema";
+
+export async function sendRequest(requestConfig) {
+  const url = `${clientEnv.NEXT_PUBLIC_BASE_URL}${requestConfig.url}`;
+  try {
+    const res = await fetch(url, {
```
lukeecart, line 6: Could we perhaps have a mapper function on this file for the Request options?

Something like:
GetRequestOptions(requestConfig) {
// run all the logic in here instead of adding it in the sendRequest method?
}

services/utils/apiRequests.js
```diff
@@ -0,0 +1,32 @@
...
+    }
+    return data;
+  } catch (err) {
+    console.error(err);
+    throw err;
+  }
+}
+
+export function formatErrorMsg(err) {
+  let errorMsg = err;
```
lukeecart, line 25: Why do we need to set errorMsg here? 

Because if the error is not an object, it's returned anyway. 

So perhaps just have if (typeof err === "object") and then return the result of Object.keys.

Else return error.

Also for readability, it's sometimes better to not shorten the variable and just set the variable to "error" instead of "err"

services/utils/apiRequests.js
```diff
@@ -0,0 +1,32 @@
...
+
+export async function sendRequest(requestConfig) {
+  const url = `${clientEnv.NEXT_PUBLIC_BASE_URL}${requestConfig.url}`;
+  try {
+    const res = await fetch(url, {
+      method: requestConfig.method ? requestConfig.method : "GET",
+      headers: {
+        "Content-Type": "application/json",
+      },
+      body: requestConfig.body ? JSON.stringify(requestConfig.body) : null,
```
lukeecart, line 11: Within the mapper we can then set the body only if there is requestConfig.body instead of setting to null

lukeecart (commented):
I like the idea and left a few comments

services/utils/apiRequests.js
badrivlog: Certainly, I appreciate your idea 👍

services/utils/apiRequests.js
badrivlog: Yes, I agree with you👍

services/utils/apiRequests.js
badrivlog: Great idea 👍

badrivlog:
> I like the idea and left a few comments

Thanks for taking the time to review👍

badrivlog:
@lukeecart,

Addressed the suggested changes and made updates to the PR👍 Could you please take another look and let me know if everything aligns with your expectations? Your feedback is important to me

lukeecart (approved):
Lgtm

eddiejaoude:
> @eddiejaoude, I believe we're now ready to proceed with merging the PR into the main codebase. Could you please take a look at the changes and, if everything looks good to you, give the final approval for merging?

Thank you @badrivlog for your efforts on this, but after discussing with the other maintainers, we feel it it extra abstraction that is not required, so I will close for now. Sorry for letting this go so far but I was trying to understand the benefit.

eddiejaoude:
Thank you for understanding @badrivlog - yes please continue contributing to LinkFree, I look forward to geeking out more 💪

PR #8311, opened 2023-07-22
chore: optimize docker development

Pradumnasaraf:
## Changes proposed

- Change script name
- Optimized the RUN command in Dockerfile
     - Now we don't need to check `development` ARG in Dockerfile because we now have separate files for prod (GHCR) and local dev.

<a href="https://gitpod.io/#https://github.com/EddieHubCommunity/LinkFree/pull/8311"></a>


eddiejaoude (approved):
Looks great 👍

PR #8380, opened 2023-07-28
fix: date format helper

### LINKED ISSUE #7650: Add date formatting functionality
### Description

To avoid repetition, either add a date formatting helper function or add a date library such as date fns.

### Screenshots

_No response_

### Additional information

_No response_

eddiejaoude:
## Fixes Issue

fixes #7650 

## Note to reviewers

<a href="https://gitpod.io/#https://github.com/EddieHubCommunity/LinkFree/pull/8380"></a>

### DIFF
--- a/pages/account/statistics.js
+++ b/pages/account/statistics.js
@@ -13,6 +13,7 @@ import PageHead from "@components/PageHead";
 import { abbreviateNumber } from "@services/utils/abbreviateNumbers";
 import Navigation from "@components/account/manage/Navigation";
 import UserMini from "@components/user/UserMini";
+import dateFormat from "@services/utils/dateFormat";
 
 const DynamicChart = dynamic(
   () => import("../../components/statistics/StatsChart"),
@@ -95,15 +96,10 @@ export async function getServerSideProps(context) {
 }
 
 export default function Statistics({ data, profile, progress, BASE_URL }) {
-  const dateTimeStyle = {
-    dateStyle: "short",
-  };
   const dailyViews = data.profile.daily.slice(-30).map((day) => {
     return {
       views: day.views,
-      date: new Intl.DateTimeFormat("en-GB", dateTimeStyle).format(
-        new Date(day.date)
-      ),
+      date: dateFormat({ format: "short", date: day.date }),
     };
   });
 
--- a/pages/account/statistics.js
+++ b/pages/account/statistics.js
@@ -85,6 +85,13 @@ export async function getServerSideProps(context) {
   }, 0);
   data.links.clicks = totalClicks;
 
+  data.profile.daily = data.profile.daily.slice(-30).map((day) => {
+    return {
+      views: day.views,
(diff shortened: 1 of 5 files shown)


services/utils/dateFormat.js
```diff
@@ -0,0 +1,17 @@
...
+      break;
+    case "long":
+      dateTimeStyle = {
+        dateStyle: "full",
+        timeStyle: "long",
+      };
+      break;
+  }
+
+  return new Intl.DateTimeFormat("en-GB", dateTimeStyle).format(new Date(date));
```
dan-mba, line 16: This will confuse people as dates will be displayed as `DD/MM/YYYY` everywhere.

eddiejaoude: But isn't that what we have at the moment? I just abstracted the duplicate code

dan-mba: I might be, but it shouldn't.

On the server side, there should be a consistent format.
But on the client side, it should use the local format.
For example in components/event/EventCard.js, the code in the `useEffect` should use local format.

With the function this way, there is no option to use local time.
To specify to use local time, the function should look like:
```js
Intl.DateTimeFormat(undefined, options)
```

eddiejaoude: Yep makes sense 👍  I will update

pages/account/statistics.js
```diff
@@ -95,15 +96,10 @@ export async function getServerSideProps(context) {
...
-  const dateTimeStyle = {
-    dateStyle: "short",
-  };
   const dailyViews = data.profile.daily.slice(-30).map((day) => {
     return {
       views: day.views,
-      date: new Intl.DateTimeFormat("en-GB", dateTimeStyle).format(
-        new Date(day.date)
-      ),
+      date: dateFormat({ format: "short", date: day.date }),
```
dan-mba, line 102: This code will have hydration issues.

When SSR is run, the locale will be "en-GB" (using Vercel in London).
On the client, it will be local (ex "en-US"), which if different, will force a complete re-render.

The way to avoid this, is to do an initial conversion in getServerSideProps. Store the value in a state variable. Then update on render with a useEffect.

eddiejaoude: Oh wow ok, I didn't realise that, so that problem exists already?

dan-mba: No.

It is caused by the switch from "en-GB" to local.

eddiejaoude: Oh I see! Thank you!

I tried to do the dynamic way with useEffect, but had some weird issues. I will put it back to `en-GB` and further improvements can come later

eddiejaoude: Pushed, hopefully this is ok now?

dan-mba: Looks good for this PR. I will look into making some improvements in a future PR.

eddiejaoude: Great! Thank you Dan ❤️

Issue #8169, opened 2023-07-13
[IMPROVEMENT] search api full-text-search


badrivlog:
### What would you like to share?

Currently, we have been filter users by JavaScript, that's not efficient for performance. Instead it would be better to perform that logic by MongoDB full-text-search.

eddiejaoude:
This would be a great improvement 👍

ArnabChatterjee20k:
Can I work on it

SaraJaoude:
> Can I work on it

@ArnabChatterjee20k I have assigned this to you

ArnabChatterjee20k:
One question I have, full text search wont search for substrings in the slug. So we have to enter the full name to get the results.
So will it be be good to implement this?

eddiejaoude:
Can you five an example of what you mean? But if I think I understand, partial search keyword match should return results

ArnabChatterjee20k:
**Full text search example**
My name - ArnabChatterjee20k
Now if I search for Arnab or arnab , it will not return results
I have to search for ArnabChatterjee20k to get the result

Partial searching or pattern searching will give the result by getting the pattern

But the issue is asking for full text search.
So I asked this question

ArnabChatterjee20k:
> Can you five an example of what you mean? But if I think I understand, partial search keyword match should return results

Yes partial or pattern searching is the ideal one here

eddiejaoude:
Yes fulltext search is designed for partial matches, similar to google search, here from the mongo docs...

> Full-text search refers to searching some text inside extensive text data stored electronically and returning results that contain some or all of the words from the query. In contrast, traditional search would return exact matches.

https://www.mongodb.com/basics/full-text-search

ArnabChatterjee20k:
I am currently facing this issue and not getting any solution. I am trying it from the last 2days. 
Here is solution and the approch and exactly what issue is coming
https://stackoverflow.com/questions/76773218/constructing-a-mongodb-query-to-emulate-javascript-filtering-logic-filterevery?noredirect=1#comment135348552_76773218

eddiejaoude:
> I am currently facing this issue and not getting any solution. I am trying it from the last 2days.
> Here is solution and the approch and exactly what issue is coming
> https://stackoverflow.com/questions/76773218/constructing-a-mongodb-query-to-emulate-javascript-filtering-logic-filterevery?noredirect=1#comment135348552_76773218

[...]

PR #8400, opened 2023-07-29
Improvement/search api#8169

ArnabChatterjee20k:
Imrproved #8169 ([IMPROVEMENT] search api full-text-search)

## Previous Workflow
* Previously the searching was getting done by vanilla js methods. 
* All the profiles were fetched then they were getting looped to filter

## Current Workflow
* Full text search based on based indices

## Dependency to setup full text search
* Enable indices in the atlas 
* Docs link https://www.mongodb.com/developer/products/atlas/build-movie-search-application/#step-4--create-a-rest-api

## Screenrecordings
* Uploading youtube video as the .mkv file is not supported by github
https://youtu.be/Zd7q6cdLlS8

<a href="https://gitpod.io/#https://github.com/EddieHubCommunity/LinkFree/pull/8400"></a>


ArnabChatterjee20k:
The search index needed to be created manually cause according to the docs [mongo atlas docs](https://www.mongodb.com/docs/atlas/atlas-search/create-index/) , the mongo search index is not available in mongoose.
This is the search index 
```json
{
  "mappings": {
    "dynamic": true
  },
  "storedSource": {
    "include": [
      "username",
      "name",
      "tags"
    ]
  }
```
For reference the docs link -> [Mongo Atlas Search Docs](https://www.mongodb.com/docs/atlas/atlas-search/create-index/)

config/searchConfig.js
```diff
@@ -0,0 +1,40 @@
+const fields = ["name", "username", "tags", "location.name"];
+const defaultPath = "name"
+function getQueryString(queryArr) {
+    console.log(queryArr);
```
adityaraute, line 4: I wonder if we should let console.log statements in production. What do the maintainers think about this?

ArnabChatterjee20k: Shall I remove the console logs? I kept them to analyze if the aggregation is working correctly or not

adityaraute: Even I'm a 100 on this. Would help if a maintainer lets us know?

eddiejaoude: Good spot, we should not have console log, we have a logger if things need to be logged

ArnabChatterjee20k: Removed the logs
Thank you aditya for spotting it and thank you eddie for the final answer

eddiejaoude: Great collaboration everyone ❤️ 💪

pages/api/search/index.js
```diff
@@ -22,39 +22,25 @@ export default async function handler(req, res) {
...
-    });
+    const filteredUsers = await searchUsers(terms);
 
     if (!filteredUsers.length) {
       return res.status(404).json({ error: `${cleanedSlug} not found` });
     }
 
     res.status(200).json({ users: filteredUsers });
   } catch (e) {
+    if (e.code === 40324) {
```
azhechkova, line 36: It's preferable not to hardcode the values in the code; instead, you can move them to constants.
f.e. 
`res.status(OKAY_STATUS_CODE);`
`if (e.code === AUTOCOMPLETE_ERROR_CODE) `

pages/api/profiles/index.js
```diff
@@ -33,3 +34,62 @@ export async function getProfiles(options = {}) {
 
   return JSON.parse(JSON.stringify(profiles));
 }
+
+export async function getUsers(options = {}) {
+  await connectMongo();
+
+  let profiles = [];
+  const { cards } = { cards: false, ...options };
```
azhechkova, line 42: I'm not sure about this. If you intend to provide a default value for the 'cards' property when it's not provided, you can also use the following syntax: `const { cards = false } = options`. In my opinion, this approach is more readable. Up to you

pages/api/profiles/index.js
```diff
@@ -33,3 +34,62 @@ export async function getProfiles(options = {}) {
...
+    profiles = await Profile.find(
+      { name: { $exists: true }, isEnabled: true },
+      fields
+    );
+  } catch (e) {
+    logger.error(e, "failed loading profiles");
+    return profiles;
+  }
+
+  return JSON.parse(JSON.stringify(profiles));
```
azhechkova, line 56: Is `JSON.parse(JSON.stringify())` used for deep cloning in this case? If not, I'm unsure why we need to first stringify the array and then parse it back

pages/api/search/index.js
```diff
@@ -1,12 +1,12 @@
...
 import connectMongo from "@config/mongo";
-import { getProfiles } from "../profiles";
 import logger from "@config/logger";
 
 export default async function handler(req, res) {
   await connectMongo();
   const { slug } = req.query;
 
-  if (req.method != "GET") {
+  if (req.method !== "GET") {
```
azhechkova, line 9: You can move methods to constants

ArnabChatterjee20k: I haven't done this cause I worked with the mongo queries only. 
I think while pulled the changes to remove conflicts it somehow get added

pages/api/profiles/index.js
eddiejaoude: There was a reason, but I can't remember why 🤔  @dan-mba do you remember why?

dan-mba: All values passed from `getServerSideProps` or `getStaticProps` to a page need to be converted to JSON.

If we let Next handle this, it will fail on certain data types returned from Mongo.

By using this hack, we return an object, but in a format that Next can successfully pass along as JSON.

eddiejaoude: Thank you Dan for explaining 👍

pages/api/search/index.js
eddiejaoude: I think this is a great suggestion 👍

config/searchConfig.js
```diff
@@ -0,0 +1,39 @@
...
+        name: 1,
+        username: 1,
+        "location.name": 1,
+        tags: 1,
+        score: { $meta: "searchScore" },
+      },
+    },
+  ];
+  return agg;
+}
```
eddiejaoude, line 39: Is this used in multiple places? I am not sure it needs to be in the config, it looks like code to me

eddiejaoude (changes requested):
Looks great, there are some inline comments, thank you 👍

config/searchConfig.js
ArnabChatterjee20k: No it is used in one place only. But since it is search query putting it with the search function can make it a bit messy. Since it is query string which is also a somekind of config , so I thought to put all these functions in a separate file searchConfig.js

eddiejaoude: Sorry for my slow reply. I think this is more confusing as people would not look there for it. Best to keep it in the actual place for now and it can be refactored/optimised later if required

ArnabChatterjee20k: ok I am implementing the changes required

eddiejaoude: Thank you 👍

ChinmayMhatre:
@ArnabChatterjee20k Do you need any help with the suggested changes?

ArnabChatterjee20k:
> @ArnabChatterjee20k Do you need any help with the suggested changes?

I need one help
https://github.com/EddieHubCommunity/BioDrop/actions/runs/6291665255/job/17080290013?pr=8400
I kept the logic same
The first time when I was merging all the test cases passed
But this time getting the error in the search
I took the aggregations from config and put them in search api as per the changes requested

Dun-sin:
> > @ArnabChatterjee20k Do you need any help with the suggested changes?
> 
> I need one help https://github.com/EddieHubCommunity/BioDrop/actions/runs/6291665255/job/17080290013?pr=8400 I kept the logic same The first time when I was merging all the test cases passed But this time getting the error in the search I took the aggregations from config and put them in search api as per the changes requested

don't think it has to do with you much, what do you think @eddiejaoude 
image

ArnabChatterjee20k:
> but some of t

I think the test cases were updated cause in the previous prs it was not giving any issues. I think the result getting returned by partial full text searching of atlas is different than the result returned by filter of native javascript.

eddiejaoude:
let me take a look, the tests might need updating

eddiejaoude:
Yes the tests need updating now the the search is returning different results]

Screenshot 2023-09-28 at 09 12 53

eddiejaoude (changes requested):
Please remove the changes to the lock file, these are to required if there are no changes to the package file

ArnabChatterjee20k:
> Please remove the changes to the lock file, these are to required if there are no changes to the package file

Reverted package.lock
since the branch got some more commit so I did a soft reset then reverted that
any better approach?

ArnabChatterjee20k:
> Yes the tests need updating now the the search is returning different results]
> 
> Screenshot 2023-09-28 at 09 12 53

@eddiejaoude @Dun-sin So any suggestion of approach that I should take cause the search and test results are not matching?

eddiejaoude:
As just discussed in the issue, having code that is different for local dev and production is not a good idea - until we can decide how to handle features of Mongo Atlas locally we will close the issue and PR. Thank you for your efforts on this 💪

PR #8524, opened 2023-08-05
Adjusted padding between section and pages

debadityasingh8972:
## Changes proposed
Adjusted Gaps between Horizontal line and hover line of pages

## Note to reviewers

<a href="https://gitpod.io/#https://github.com/EddieHubCommunity/LinkFree/pull/8524"></a>

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -6,7 +6,7 @@
   "packages": {
     "": {
       "name": "linkfree",
-      "version": "1.168.6",
+      "version": "1.160.3",
       "hasInstallScript": true,
       "license": "MIT",
       "dependencies": {
--- a/package-lock.json
+++ b/package-lock.json
@@ -6,7 +6,7 @@
   "packages": {
     "": {
       "name": "linkfree",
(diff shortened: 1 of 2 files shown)


adityaraute (changes requested):
Might be a good idea to revert changes for the package-lock.json file as it should remain constant for the repo. I can see the rest of your code isn't using any additional modules so this shouldn't cause an impact.

adityaraute:
Also, I don't think this is ideal.

If you look at the first row - there is a lot of padding at the top and not enough at the bottom of the card
image


But on the second row - there is the same amount of padding at the top but a lot of padding at the bottom (because of the card on the right I assume)
image

Also a lot of gap between the two rows (that could be utilized by the top row)

It's not a showstopper, But I feel it can be modified

debadityasingh8972:
Okay. I'm working on it.

amandamartin-dev (approved):
Checked in gitpod, looks good to me. Thank you for your contribution!

PR #8605, opened 2023-08-10
feat: log profile updates that occur by the manage forms

### LINKED ISSUE #7660: [FEATURE] Log all database changes
### Description

Each time there is a db write/update, have a log collection with the following info

- username
- collection
- collection id
- changes before
- changes after
- date

This can be achieved with the `pre` middleware hook on the models https://mongoosejs.com/docs/middleware.html

### Screenshots

_No response_

### Additional information

_No response_

babblebey:
This Pull Request Introduces a new model `Changelog` to the project which keeps records of changes that occurs in the `Link` and `Profile` models.

## Fixes Issue

Closes #7660 

## Changes proposed

- Add a new model `Changelog` with the following schema

```js
// models/Changelog.js

{
  username: String,
  docId: mongoose.Schema.Types.ObjectId,
  collectionName: String,
  changesBefore: mongoose.Schema.Types.Mixed,
  changesAfter: mongoose.Schema.Types.Mixed,
  diff: mongoose.Schema.Types.Mixed,
},
{ timestamps: true }
```

- Implement a `dbChangesLoggerMiddleware` function that attaches the necessary middleware to model's `schema`, computes value for each of the `Changelog` model property to create a `change` object, which it in turn saves to the `changelogs` database collection and logs to the terminal.

```js
// models/middlewares/dbChangesLogger.js

export default function dbChangesLoggerMiddleware(schema) {
  let isUpdateOperation;

  let username;
  let collection;
  let changesBefore;
  let changesAfter;
  let diff;

  async function beforeUpdate(next) {
    isUpdateOperation = this.op && this.op === "findOneAndUpdate";

    username = isUpdateOperation 
      ? this._conditions.username 
      : this.username;
    collection = isUpdateOperation 
      ? this._collection.collectionName 
      : this.constructor.collection.collectionName;
    changesBefore = isUpdateOperation 
      ? await this.model.findOne(this.getFilter()) 
      : null;
  
    next();
  }

  async function afterUpdate(doc, next) {
    changesAfter = await doc.constructor.findById(doc._id);

    if (isUpdateOperation && changesBefore && changesAfter) {
      diff = getChangesDiff(changesBefore._doc, changesAfter._doc);
    } else {
      diff = doc;
    }

    const change = {
      username,
      collectionName: collection,
      docId: doc._id,
      changesBefore,
      changesAfter,
      diff,
    };

    if (changesBefore !== changesAfter) {
      await Changelog.create({ ...change });
      logger.info(change);
    } else {
      logger.info(`No changes occured in: ${collection}`)
    }

    return next();
  }
  
  schema.pre("save", beforeUpdate);
  schema.pre("findOneAndUpdate", beforeUpdate);

  schema.post("save", afterUpdate);
  schema.post("findOneAndUpdate", afterUpdate);
}

function getChangesDiff(chngB, chngA) {
  const diff = {};

  for (const key in chngA) {
    if (JSON.stringify(chngB[key]) !== JSON.stringify(chngA[key])) {
      diff[key] = chngA[key];
    }
  }

  return diff;
}
```

- Integrate `dbChangesLoggerMiddleware` function into the `Link` and `Profile` models to track and keep changes that occurs in them, by adding the below line before the `model` is created after the `schema`

```js
// model/Link.js

dbChangesLoggerMiddleware(LinkSchema);
```

```js
// model/Profile.js

dbChangesLoggerMiddleware(ProfileSchema);
```

## Screenshots

image

## Note to reviewers

<a href="https://gitpod.io/#https://github.com/EddieHubCommunity/LinkFree/pull/8605"></a>


utils/functions/dbChangesLogger.js
```diff
@@ -0,0 +1,49 @@
+import logger from "@config/logger";
```
eddiejaoude, line 1: Please update your branch, because the `utils` directory has been removed. I suggest moving this re-usable function to the `models` directory as it will only be used in there (maybe to `model/index.js` for now)

babblebey: So, I moved the function to a new directory at `models/middlewares` instead, I thought it wouldn't be nice to let the `models/index.js` deviate from its simple functionality of group exporting the models.

> ...the `models` directory as it will only be used in there...

This way we are making sure the `dbChangesLoggerMiddleware` is not in any way made available to other file during `import` from `models` since it will only be used by `models`

```js
import { dbChangesLoggerMiddleware } from "@models/index";
```

If it will be imported for any reason, then it must be intentional

```js
import dbChangesLoggerMiddleware from "@models/middlewares/dbChangesLogger";
```

eddiejaoude (changes requested):
Looking good 👍  

Please can you create a new model to store the data

utils/functions/dbChangesLogger.js
eddiejaoude: Ok thank you 👍

models/middlewares/dbChangesLogger.js
```diff
@@ -0,0 +1,49 @@
...
+      username,
+      collection,
+      collection_id: "in progress....",
+      changesBefore,
+      changesAfter,
+      date
+    };
+    
+    if (changesBefore !== changesAfter) {
+      logger.info(change);
```
eddiejaoude, line 34: I am keen to log these changes to a new collection in Mongo

babblebey: Same here Eddie! 

But I have a certain blocker... Its weird. So..

- I have consumed the `mongoose` doc enough to know that I should to write a `Schema` then create a `Model` from the schema like so.

```js
const myNewSchema = new mongoose.Schema({ ... });

mongoose.model('MyModel', myNewSchema);
```
- I have also taken notice of style the existing models in the project are created and I have arrived at my implementation in similar fashion. _**Below is my implementation of the `DBChange` model**_ (Blocker Here)

```js
 
const dbChangeSchema = new mongoose.Schema(
  {
    username: String,
    collection: String,
    doc: String,
    changesBefore: mongoose.Schema.Types.Mixed,
    changesAfter: mongoose.Schema.Types.Mixed,
    date: {
      type: Date,
      default: new Date(),
    }
  },
  { timestamps: true }
);

module.exports = mongoose.models.DBChange || mongoose.model("DBChange", dbChangeSchema);
```
- I hope to implement this in the `dbChangesLogger` middleware function

```js
  async function afterUpdate(doc, next) {
    ...
    const change = {
      username,
      collection,
      doc: doc._id,
      changesBefore,
      changesAfter,
      date
    };
    
    if (changesBefore !== changesAfter) {
      await DBChange.create({ ...change });   // <- CHANGE ADDED TO DB RIGHT HERE
      logger.info(change);
    } else {
      logger.info(`No changes occured in: ${collection}`)
    }

    return next();
  }
  ...
};
```

...but there goes my certain blocker.

### Blocker

- I get the below error when I try to load the website on 

**Screenshot**

screenshot-localhost_3000-2023 08 14-20_51_55

**Stack Trace**
```shell
(node:2716) [MONGOOSE] Warning: `collection` is a reserved schema pathname and may break some functionality. You are allowed to use it, but use at your own risk. To disable this warning pass `suppressReservedKeysWarning` as a schema option.
(Use `node --trace-warnings ...` to show where the warning was created)
- error Error [TypeError]: Cannot read properties of undefined (reading 'Symbol(mongoose#Document#scope)')
    at Model.set [as collection] (C:\Users\Bode\Documents\Projects\Open Source\LinkFree\node_modules\mongoose\lib\helpers\document\compile.js:205:32)
    at Function.compile (C:\Users\Bode\Documents\Projects\Open Source\LinkFree\node_modules\mongoose\lib\model.js:4735:30)
    at Mongoose._model (C:\Users\Bode\Documents\Projects\Open Source\LinkFree\node_modules\mongoose\lib\index.js:620:27)
    at Mongoose.model (C:\Users\Bode\Documents\Projects\Open Source\LinkFree\node_modules\mongoose\lib\index.js:579:27)
    at eval (webpack-internal:///./models/DBChange.js:20:73)
    at ./models/DBChange.js (C:\Users\Bode\Documents\Projects\Open Source\LinkFree\.next\server\pages\account\manage\links.js:319:1)
    at __webpack_require__ (C:\Users\Bode\Documents\Projects\Open Source\LinkFree\.next\server\webpack-runtime.js:33:42)
    at eval (webpack-internal:///./models/index.js:16:67)
    at ./models/index.js (C:\Users\Bode\Documents\Projects\Open Source\LinkFree\.next\server\pages\account\manage\links.js:451:1)
    at __webpack_require__ (C:\Users\Bode\Documents\Projects\Open Source\LinkFree\.next\server\webpack-runtime.js:33:42)
    at eval (webpack-internal:///./pages/api/auth/db-adapter.js:5:71)
    at ./pages/api/auth/db-adapter.js (C:\Users\Bode\Documents\Projects\Open Source\LinkFree\.next\server\pages\account\manage\links.js:704:1)
    at __webpack_require__ (C:\Users\Bode\Documents\Projects\Open Source\LinkFree\.next\server\webpack-runtime.js:33:42)
    at eval (webpack-internal:///./pages/api/auth/[...nextauth].js:12:69)
    at ./pages/api/auth/[...nextauth].js (C:\Users\Bode\Documents\Projects\Open Source\LinkFree\.next\server\pages\account\manage\links.js:693:1)
    at __webpack_require__ (C:\Users\Bode\Documents\Projects\Open Source\LinkFree\.next\server\webpack-runtime.js:33:42)
    at eval (webpack-internal:///./pages/account/manage/links.js:9:77)
    at __webpack_require__.a (C:\Users\Bode\Documents\Projects\Open Source\LinkFree\.next\server\webpack-runtime.js:97:13)
    at eval (webpack-internal:///./pages/account/manage/links.js:1:21)
    at ./pages/account/manage/links.js (C:\Users\Bode\Documents\Projects\Open Source\LinkFree\.next\server\pages\account\manage\links.js:660:1)
    at __webpack_require__ (C:\Users\Bode\Documents\Projects\Open Source\LinkFree\.next\server\webpack-runtime.js:33:42)
    at eval (webpack-internal:///./node_modules/next/dist/build/webpack/loaders/next-route-loader/index.js?page=%2Faccount%2Fmanage%2Flinks&preferredRegion=&absolutePagePath=.%2Fpages%5Caccount%5Cmanage%5Clinks.js&absoluteAppPath=private-next-pages%2F_app&absoluteDocumentPath=private-next-pages%2F_document&middlewareConfigBase64=e30%3D!:23:88)
    at __webpack_require__.a (C:\Users\Bode\Documents\Projects\Open Source\LinkFree\.next\server\webpack-runtime.js:97:13)
    at eval (webpack-internal:///./node_modules/next/dist/build/webpack/loaders/next-route-loader/index.js?page=%2Faccount%2Fmanage%2Flinks&preferredRegion=&absolutePagePath=.%2Fpages%5Caccount%5Cmanage%5Clinks.js&absoluteAppPath=private-next-pages%2F_app&absoluteDocumentPath=private-next-pages%2F_document&middlewareConfigBase64=e30%3D!:1:21)
    at ./node_modules/next/dist/build/webpack/loaders/next-route-loader/index.js?page=%2Faccount%2Fmanage%2Flinks&preferredRegion=&absolutePagePath=.%2Fpages%5Caccount%5Cmanage%5Clinks.js&absoluteAppPath=private-next-pages%2F_app&absoluteDocumentPath=private-next-pages%2F_document&middlewareConfigBase64=e30%3D! (C:\Users\Bode\Documents\Projects\Open Source\LinkFree\.next\server\pages\account\manage\links.js:33:1)
    at __webpack_require__ (C:\Users\Bode\Documents\Projects\Open Source\LinkFree\.next\server\webpack-runtime.js:33:42)
    at __webpack_exec__ (C:\Users\Bode\Documents\Projects\Open Source\LinkFree\.next\server\pages\account\manage\links.js:1546:39)
    at C:\Users\Bode\Documents\Projects\Open Source\LinkFree\.next\server\pages\account\manage\links.js:1547:28
    at Object.<anonymous> (C:\Users\Bode\Documents\Projects\Open Source\LinkFree\.next\server\pages\account\manage\links.js:1550:3)
    at Module._compile (node:internal/modules/cjs/loader:1256:14)
    at Module._extensions..js (node:internal/modules/cjs/loader:1310:10)
    at Module.load (node:internal/modules/cjs/loader:1119:32)
    at Module._load (node:internal/modules/cjs/loader:960:12)
    at Module.require (node:internal/modules/cjs/loader:1143:19)
    at require (node:internal/modules/cjs/helpers:110:18)
    at requirePage (C:\Users\Bode\Documents\Projects\Open Source\LinkFree\node_modules\next\dist\server\require.js:112:75)
    at C:\Users\Bode\Documents\Projects\Open Source\LinkFree\node_modules\next\dist\server\load-components.js:80:84
    at async loadComponentsImpl (C:\Users\Bode\Documents\Projects\Open Source\LinkFree\node_modules\next\dist\server\load-components.js:80:26)
    at async DevServer.findPageComponentsImpl (C:\Users\Bode\Documents\Projects\Open Source\LinkFree\node_modules\next\dist\server\next-server.js:772:36)
    at async DevServer.findPageComponents (C:\Users\Bode\Documents\Projects\Open Source\LinkFree\node_modules\next\dist\server\dev\next-dev-server.js:1271:20)
    at async DevServer.renderPageComponent (C:\Users\Bode\Documents\Projects\Open Source\LinkFree\node_modules\next\dist\server\base-server.js:1357:24)
    at async DevServer.renderToResponseImpl (C:\Users\Bode\Documents\Projects\Open Source\LinkFree\node_modules\next\dist\server\base-server.js:1401:32)
    at async DevServer.pipeImpl (C:\Users\Bode\Documents\Projects\Open Source\LinkFree\node_modules\next\dist\server\base-server.js:645:25)
    at async Object.fn (C:\Users\Bode\Documents\Projects\Open Source\LinkFree\node_modules\next\dist\server\next-server.js:1153:21)
    at async Router.execute (C:\Users\Bode\Documents\Projects\Open Source\LinkFree\node_modules\next\dist\server\router.js:315:32)
    at async DevServer.runImpl (C:\Users\Bode\Documents\Projects\Open Source\LinkFree\node_modules\next\dist\server\base-server.js:619:29)
    at async DevServer.run (C:\Users\Bode\Documents\Projects\Open Source\LinkFree\node_modules\next\dist\server\dev\next-dev-server.js:910:20)
    at async DevServer.handleRequestImpl (C:\Users\Bode\Documents\Projects\Open Source\LinkFree\node_modules\next\dist\server\base-server.js:546:20) {
  digest: undefined
}
```

### Other concerns

- We have some sort of `warning` in the stack trace about our usage of `collection` as property name

```shell
Warning: `collection` is a reserved schema pathname and may break some functionality. You are allowed to use it, but use at your own risk. To disable this warning pass `suppressReservedKeysWarning` as a schema option.
(Use `node --trace-warnings ...` to show where the warning was created)
```

eddiejaoude: Can you push it to your branch so I can test it locally and try to debug

babblebey: Pushed!

eddiejaoude: I made some tweaks and it then worked for me, here are the changes, save them to a file `8605.patch` and you can do `git apply 8605.patch`

```diff
diff --git a/models/Changelog.js b/models/Changelog.js
new file mode 100644
index 00000000..e3adb591
--- /dev/null
+++ b/models/Changelog.js
@@ -0,0 +1,15 @@
+import mongoose from "mongoose";
+
+const ChangelogSchema = new mongoose.Schema(
+  {
+    username: String,
+    collectionId: String,
+    collectionName: String,
+    changesBefore: mongoose.Schema.Types.Mixed,
+    changesAfter: mongoose.Schema.Types.Mixed,
+  },
+  { timestamps: true }
+);
+
+module.exports =
+  mongoose.models.Changelog || mongoose.model("Changelog", ChangelogSchema);
diff --git a/models/DBChange.js b/models/DBChange.js
deleted file mode 100644
index 3dd4c487..00000000
--- a/models/DBChange.js
+++ /dev/null
@@ -1,18 +0,0 @@
-import mongoose from "mongoose";
- 
-const dbChangeSchema = new mongoose.Schema(
-  {
-    username: String,
-    collection: String,
-    doc: String,
-    changesBefore: mongoose.Schema.Types.Mixed,
-    changesAfter: mongoose.Schema.Types.Mixed,
-    date: {
-      type: Date,
-      default: new Date(),
-    }
-  },
-  { timestamps: true }
-);
-
-module.exports = mongoose.models.DBChange || mongoose.model("DBChange", dbChangeSchema);
\ No newline at end of file
diff --git a/models/index.js b/models/index.js
index 5b9f86fa..79b4a2ff 100644
--- a/models/index.js
+++ b/models/index.js
@@ -1,5 +1,5 @@
 import Account from "./Account";
-import DBChange from "./DBChange";
+import Changelog from "./Changelog";
 import Link from "./Link";
 import LinkStats from "./LinkStats";
 import Profile from "./Profile";
@@ -11,7 +11,7 @@ import VerificationToken from "./VerificationToken";
 
 export {
   Account,
-  DBChange,
+  Changelog,
   Link,
   LinkStats,
   Profile,
diff --git a/models/middlewares/dbChangesLogger.js b/models/middlewares/dbChangesLogger.js
index 0ceba02c..0cf1b86d 100644
--- a/models/middlewares/dbChangesLogger.js
+++ b/models/middlewares/dbChangesLogger.js
@@ -1,5 +1,5 @@
 import logger from "@config/logger";
-import { DBChange } from "@models/index";
+import { Changelog } from "@models/index";
 
 const dbChangesLoggerMiddleware = (schema) => {
   let username;
@@ -11,9 +11,13 @@ const dbChangesLoggerMiddleware = (schema) => {
     const isUpdateOperation = this.op && this.op === "findOneAndUpdate";
 
     username = isUpdateOperation ? this._conditions.username : this.username;
-    collection = isUpdateOperation ? this._collection.collectionName : this.constructor.collection.collectionName;
-    changesBefore = isUpdateOperation ? await this.model.findOne(this.getFilter()) : null;
-  
+    collection = isUpdateOperation
+      ? this._collection.collectionName
+      : this.constructor.collection.collectionName;
+    changesBefore = isUpdateOperation
+      ? await this.model.findOne(this.getFilter())
+      : null;
+
     next();
   }
 
@@ -21,26 +25,25 @@ const dbChangesLoggerMiddleware = (schema) => {
     changesAfter = await doc.constructor.findById(doc._id);
 
     const date = new Date();
-    
+
     const change = {
       username,
-      collection,
-      doc: doc._id,
+      collectionName: collection,
+      collectionId: doc._id,
       changesBefore,
       changesAfter,
-      date
     };
-    
+
     if (changesBefore !== changesAfter) {
-      await DBChange.create({ ...change });
+      await Changelog.create({ ...change });
       logger.info(change);
     } else {
-      logger.info(`No changes occured in: ${collection}`)
+      logger.info(`No changes occured in: ${collection}`);
     }
 
     return next();
   }
-  
+
   schema.pre("save", beforeUpdate);
   schema.pre("findOneAndUpdate", beforeUpdate);
```

eddiejaoude:
It is looking really good 👍 

Is it possible to have a field in the model to show the diff for easier viewing?

babblebey:
> Is it possible to have a field in the model to show the diff for easier viewing?

Just the difference between the `changesBefore` and `changesAfter`!? I'll look into that ✌🏾

eddiejaoude:
> > Is it possible to have a field in the model to show the diff for easier viewing?
> 
> Just the difference between the `changesBefore` and `changesAfter`!? I'll look into that ✌🏾

That would be great! I think something like...

BEFORE

```js
// ...
bio: "aaaa"
// ...
```

AFTER

```js
// ...
bio: "bbbb"
// ...
```

DIFF (new field)

```js
// ...
bio: "bbbb"
// ...
```

I think so only the changes fields with their new values are stored in the new `diff` property

models/middlewares/dbChangesLogger.js
babblebey: Works beautifully @eddiejaoude 

Haha, I will just take it that the `warning` about the `collection` reserved term usage indeed broke some functionality as mentioned in there,

```shell
Warning: `collection` is a reserved schema pathname and may break some functionality. You are allowed to use it, but use at your own risk. To disable this warning pass `suppressReservedKeysWarning` as a schema option.
(Use `node --trace-warnings ...` to show where the warning was created)
```

...considering there wasn't lots of changes beyond the rename context. 

### Notable Learning

- Now I see what the `{ timestamps: true }` does in the schema, I see it adds the `createdAt` and `upatedAt` timestamp properties.

### Other Thoughts

- I see that you have opted for the term `collectionId` to hold the `doc._id`. I had some thoughts concerning the property naming... _see comment_ https://github.com/EddieHubCommunity/LinkFree/issues/7660#issuecomment-1675863310

😉 PS: I wasn't able to get the changes apply via `git apply <.patch file>` command due to some `trailing whitespace.` and `corrupt patch at line***` error. I tried to fix the patch using the `--whitespace=fix` flag, but no error persists, I had to manually commit the changes.

babblebey:
Exactly what I had in mind @eddiejaoude , so here's what I've got...

I tried 3 options... 1 failed❌, 1 tried❔ , 1 worked✅! 

I will start with the one that worked with its caveat and I want your confirmation before I commit this!

### Option 1 - Implementation Using `lodash` ✅

```js
if (isUpdateOperation) {
  const { _doc, ...docDiffs } = _.omitBy(changesAfter, (value, key) => _.isEqual(value, changesBefore[key]));
  diff = docDiffs;
} else {
  diff = doc;
}
```

- I was able to compute my differences using the `omitBy()` function to remove all properties in `changeAfter` that are equal to properties in `changeBefore` (using the `isEqual()` function)
- I realise that was a return of the original document copy too in a `_doc` property with the differences, so I exclude '_doc' property and spread other differences properties into a new object `docDiffs` which I inturn assigned as value for our all new `diff` Changelog model property
- It is important to state the differences detection operation will only run when `isUpdateOperation` because `changeBefore` at this point will be `null` so we return the `doc` as value for `diff`.

**Result - Desired ✅**

```shell
username: "babblebey"
collectionName: "links"
collectionId: "64d385107a9cd8bca6a08bbe"
changesBefore: {
   ...
  "name": "HRMOffice Website",
  "url": "https://hrmoffice.org",
  "icon": "FaFacebook",
  "isEnabled": true,
  "isPinned": true,
  ...
}
changesAfter: {
  ...
  "name": "HRMOfficE",
  "url": "https://hrmoffice.org",
  "icon": "FaGithub",
  "isEnabled": false,
  "isPinned": false,
  ...
}
diff: {
  "name": "HRMOfficE",
  "icon": "FaGithub",
  "isEnabled": false,
  "isPinned": false,
  "updatedAt": "2023-08-15T19:03:41.750Z"
}
```

**Caveat: This implementation requires adding a new package `lodash` to the project.**

______

### Option 2 - Implementation with custom `getChangesDiff` function ❔

```js
function getChangesDiff(chngB, chngA) {
  const diff = {};
  for (const key in chngA) {
    if (chngB[key] !== chngA[key]) {
      diff[key] = chngA[key];
    }
  }
  return diff;
}

// Usage
if (isUpdateOperation) {
  diff = getChangesDiff(changesBefore._doc, changesAfter._doc);
} else {
  diff = doc;
}
```

- I implemented the function that pretty much work the same, it add properties from `changeAfter` that are not equal to (i.e. not with same value) properties in `changeBefore` to a new object `diff`.
- Now, I had to use the `._doc` property I weirdly founding output as pointer to access the document within the `changesBefore/After`.

**Result - Not Exactly❔**

```shell
username: "babblebey"
collectionName: "links"
collectionId: "64d385107a9cd8bca6a08bbe"
changesBefore: {
   ...
  "name": "HRMOffice Website",
  "url": "https://hrmoffice.org",
  "icon": "FaFacebook",
  "isEnabled": true,
  "isPinned": true,
  ...
}
changesAfter: {
  ...
  "name": "HRMOfficE",
  "url": "https://hrmoffice.org",
  "icon": "FaGithub",
  "isEnabled": false,
  "isPinned": false,
  ...
}
diff: {
  "_id": "64d385107a9cd8bca6a08bbe", // ?? Unwanted !!
  "name": "HRMOfficE",
  "icon": "FaGithub",
  "isEnabled": false,
  "isPinned": false,
  "profile": "64ce7e0146acee92873d99ee", // ?? Unwanted !!
  "createdAt": "2023-08-09T12:22:40.424Z", // ?? Unwanted !!
  "updatedAt": "2023-08-15T19:03:41.750Z"
}
```

- The `diff` property object continues to come with the `_id`, `profile` and `createdAt` properties no matter what property is updated in the `Link` model for example, It does the same thing for the `Profile` model, there are some properties that just comes with the model.

**Note: I would've really preferred this option, because it would've been more lightweight. I assume the undesired output to be a result of some edge cases with in `getChangesDiff` func implementation, that maybe I just need to figure out.**

_______

### Option 3 - Implementation using the mongoose `getUpdate()` function ❌

```js
diff = isUpdateOperation ? (this.getUpdate()).$set : this;
```

- Pretty straightforward write in attempt to compute the difference by picking out the value from the `$set` property used in the update action `conditions`.

**Result - Undesired ❌** 

```shell
username: "babblebey"
collectionName: "links"
collectionId: "64d385107a9cd8bca6a08bbe"
changesBefore: {
   ...
  "name": "HRMOffice Website",
  "url": "https://hrmoffice.org",
  "icon": "FaFacebook",
  "isEnabled": true,
  "isPinned": true,
  ...
}
changesAfter: {
  ...
  "name": "HRMOfficE",
  "url": "https://hrmoffice.org",
  "icon": "FaGithub",
  "isEnabled": false,
  "isPinned": false,
  ...
}
diff: {
  "updatedAt": "2023-08-15T19:03:41.750Z",
  "isPinned": false,
  "isEnabled": false,
  "icon": "FaGithub",
  "url": "https://hrmoffice.org",
  "name": "HRMOfficE",
  "group": ""
}
```
- The result pretty much just carries the state of the form fields value at submission, changed or unchanged, I guess as it should.

So No Way for this Implementation!

eddiejaoude:
Wow very well thought out @babblebey 💪 I think we should try to not include lodash, as we are close to the Vercel file limit already.

I think option 2 is a great start, even if it has some extra fields - we can improve this further later in another pull request

babblebey:
Great stuff @eddiejaoude, I found a fix!

Since hearing you subscribed to (my preferred) Option 2, I said I was gonna make it work i.e. make sure it produces our desired result before in this PR, haha I can't procrastinate to another PR to make it work.

> ...I assume the undesired output to be a result of some edge cases with in `getChangesDiff` func implementation, that maybe I just need to figure out.

So here's what I've discovered as the Edge Case that the `getChangesDiff` wasn't addressing.

- The `changesBefore/After` documents are obviously of their mongoDB `Document` type, hence they maintain their `ObjectId` and some `Maps` related thing within them.

```shell
 _id: new ObjectId("64d385107a9cd8bca6a08bbe"),
username: 'babblebey',
group: '',
name: 'HRMOffice',
url: 'https://hrmoffice.org',
icon: 'FaFacebook',
isEnabled: true,
isPinned: true,
clicks: 0,
profile: new ObjectId("64ce7e0146acee92873d99ee"),
createdAt: 2023-08-09T12:22:40.424Z,
updatedAt: 2023-08-16T19:55:48.449Z,
__v: 0
```

- Meanwhile, the comparison conditions `chngB[key] !== chngA[key]` inside the `getChangesDiff` function is only able to compare the properties of type `number`, `strings` and `boolean` (I guess the primitives), leaving out `object` i.e. other types like `Date`, `Array`, `Object` etc. 

```shell
 _id: new ObjectId("64d385107a9cd8bca6a08bbe"),       // Mongo ObjectId type SKIPPED AT CHECK
username: 'babblebey',
group: '',
name: 'HRMOffice',
url: 'https://hrmoffice.org',
icon: 'FaFacebook',
isEnabled: true,
isPinned: true,
clicks: 0,
profile: new ObjectId("64ce7e0146acee92873d99ee"),       //  ObjectId type SKIPPED AT CHECK
createdAt: 2023-08-09T12:22:40.424Z,       // Date Object type SKIPPED AT CHECK
updatedAt: 2023-08-16T19:55:48.449Z,       // Date Object type SKIPPED AT CHECK - THOUGH this should make it into `diff` because it will change
__v: 0
```

...for this reason the checks for these non-primitive property types are skipped at check, hence making their way back into the `diff`. There are more types in the `Profile` model changes `doc` too, these are normal nested `Object` and `Array` object types, they all also skipped checks.

### The Fix

```js
function getChangesDiff(chngB, chngA) {
  const diff = {};
  for (const key in chngA) {
    if (JSON.stringify(chngB[key]) !== JSON.stringify(chngA[key])) {
      diff[key] = chngA[key];
    }
  }
  return diff;
}
```

- I implemented a `JSON.stringify`, the compared value of the `changes***` documents, ensuring that they are flatten fairly to `JSON` string to aid easy comparison.

Haha, `JSON` object and its methods are stuffs that `mongoose` uses quite frequently behind the hood to make `docs` readable, so I assumed that using it on our own side for the `getChangesDiff` shouldn't really cause a sweat for performance, hence I threw it in considering it got the job done.

Now, we have got a working `diff` property.

```shell
username: "babblebey"
collectionName: "links"
collectionId: "64d385107a9cd8bca6a08bbe"
changesBefore: {
  "_id": "64d385107a9cd8bca6a08bbe",
  "username": "babblebey",
  "group": "",
  "name": "HRMOffice",
  "url": "https://hrmoffice.org",
  "icon": "FaFacebook",
  "isEnabled": false,
  "isPinned": false,
  "clicks": 0,
  "profile": "64ce7e0146acee92873d99ee",
  "createdAt": "2023-08-09T12:22:40.424Z",
  "updatedAt": "2023-08-16T19:33:44.382Z",
  "__v": 0
}
changesAfter: {
  "_id": "64d385107a9cd8bca6a08bbe",
  "username": "babblebey",
  "group": "",
  "name": "HRMOffice",
  "url": "https://hrmoffice.org",
  "icon": "FaFacebook",
  "isEnabled": true,
  "isPinned": true,
  "clicks": 0,
  "profile": "64ce7e0146acee92873d99ee",
  "createdAt": "2023-08-09T12:22:40.424Z",
  "updatedAt": "2023-08-16T19:55:48.449Z",
  "__v": 0
}
diff: {
  "isEnabled": true,
  "isPinned": true,
  "updatedAt": "2023-08-16T19:55:48.449Z"
}
```

eddiejaoude:
Wow that looks great! Btw there is a build error

```
error models/Link.js (52:26) @ LinkSchema
error Error [TypeError]: (0 , _dbChangesLogger.default) is not a function
    at eval (webpack-internal:///(api)/./models/Link.js:56:30)
```

babblebey:
> Wow that looks great! Btw there is a build error
> 
> ```
> error models/Link.js (52:26) @ LinkSchema
> error Error [TypeError]: (0 , _dbChangesLogger.default) is not a function
>     at eval (webpack-internal:///(api)/./models/Link.js:56:30)
> ```

Oh, I noticed this in the Build, I thought to keep till later.

I will work on this now!

babblebey:
I have fixed the `Build / build (build) (pull_request)` @eddiejaoude with the commits above..

Now, there are 2 failing tests... which can be traced to the routes.

1. `page/search.js` - `search.spec.js` test failed justifies that no user (which accounts for the `main li`) is getting listed on the search page, though its listed in my local dev but very slow to render.

image

2. `page/discover.js` - discover.spec.js test failed justifies that no user (which accounts for the `main li`) is getting listed on the discover page, though its listed in my local dev also very slow to render.

image

Please have a look!

eddiejaoude:
Thank you for fixing those 👍  Sure I will take a look tomorrow

babblebey:
Hey @eddiejaoude 

PR is ready for review, on the last note, I just really need your input to address this concern I have in regard the `collectionId` naming.

_see comment_ https://github.com/EddieHubCommunity/LinkFree/issues/7660#issuecomment-1675863310

babblebey:
> Yes ObjectId would be better if it is possible but I don't think it is a blocker

Great, I will look to implement the `ObjectId` in the `collectionId?`. Blocker!? sure I know it shouldn't be, I just thought it'd be smooth for DX when it used straight without the need to convert types.

> Just something I thought of, can the changelog middleware ignore some updates? For example like profile views incremented?

Hmm, true.. Profile view isn't a form input update! I will first need to look at how the Profile view increments and see what I can come up with!

eddiejaoude:
> I have discovered that the Profile View action doesn't trigger the dbChangesLogger middleware function, and this is because the view property is incremented using the updateOne query/model function as implemented in the getUserApi.

> The updateOne function as said in the mongoose docs...

>Note: updateOne will not fire update middleware. Use pre('updateOne') and post('updateOne') instead.
> ...does not fire the update middleware, so our middleware findOneAndUpdate is safe off the action.

Great 🎉

eddiejaoude:
During testing I noticed that the `stats` incrementing on the `profile` does record a changelog, can we exclude this?

https://github.com/EddieHubCommunity/LinkFree/blob/605f233d2529786e3768d972f9fe484fb2576d93/pages/api/profiles/%5Busername%5D/index.js#L160-L176

babblebey:
What's the action that triggers that?

eddiejaoude:
When the profile views are increment here [LinkFree/pages/api/profiles/[username]/index.js](https://github.com/EddieHubCommunity/LinkFree/blob/605f233d2529786e3768d972f9fe484fb2576d93/pages/api/profiles/%5Busername%5D/index.js#L160-L176)

babblebey:
> When the profile views are increment here [LinkFree/pages/api/profiles/[username]/index.js](https://github.com/EddieHubCommunity/LinkFree/blob/605f233d2529786e3768d972f9fe484fb2576d93/pages/api/profiles/%5Busername%5D/index.js#L160-L176)

Hey @eddiejaoude 

I have done further investigation, I want to insist that the `views` or `stats` increment doesn't trigger the `dbChangesLogger` middleware neither does it add a `Changelog`.

I assumed a visit to a user link page, on the `/[username]` route triggers the required action in that `/api/profiles/[username]/index.js` route and I tested! That part of the code base you pointed to, seem like it's just been introduced, But the `stats` increment doesn't add a `Changelog`.

It is important to point this certain behavior to you and state that you should not confuse it with the `changelog` you might have seen recorded.

The certain behaviour is one that I have noted on the `Profile` model

- on `Link` Edit, a `Changelog` is recorded for `Profile` too after one is already recorded for `Link`, but the `diff` in their will be just the `updatedAt`

image

...this is because on add new/edit link action the `Profile` model `links` property gets updated accordingly, so in the case where its an update action, the are no new item added to the `Profile` model but the document gets update, hence the `diff` being just `updatedAt`.

If this is the `changelog` you have seen, kindly note that its not from that part of the codebase.

models/Changelog.js
```diff
@@ -0,0 +1,15 @@
...
+    collectionId: mongoose.Schema.Types.ObjectId,
+    collectionName: String,
+    changesBefore: mongoose.Schema.Types.Mixed,
+    changesAfter: mongoose.Schema.Types.Mixed,
+    diff: mongoose.Schema.Types.Mixed,
+  },
+  { timestamps: true }
+);
+
+module.exports = mongoose.models.Changelog || mongoose.model("Changelog", ChangelogSchema);
```
eddiejaoude, line 15: Extra blank like suggested by GitHub 

```suggestion
module.exports = mongoose.models.Changelog || mongoose.model("Changelog", ChangelogSchema);

```

models/middlewares/dbChangesLogger.js
```diff
@@ -0,0 +1,74 @@
...
+  const diff = {};
+
+  for (const key in chngA) {
+    if (JSON.stringify(chngB[key]) !== JSON.stringify(chngA[key])) {
+      diff[key] = chngA[key];
+    }
+  }
+
+  return diff;
+}
```
eddiejaoude, line 74:
```suggestion
}

```

models/middlewares/dbChangesLogger.js
```diff
@@ -0,0 +1,74 @@
...
+      await Changelog.create({ ...change });
+      logger.info(change);
+    } else {
+      logger.info(`No changes occured in: ${collection}`)
+    }
+
+    return next();
+  }
+  
+  schema.pre("save", beforeUpdate);
```
eddiejaoude, line 57: I wonder if the automated tests fail due to a promise issue? In the docs for the newer version of mongoose it uses the example...

```js
schema.pre('save', async function() {
  await doStuff();
  await doMoreStuff();
});
```

https://mongoosejs.com/docs/middleware.html#pre

eddiejaoude: It is just a suggestion, I am not sure why the tests file - it seems like a timeout each time and also random, which is why I think it might be a promise issue 🤔

babblebey: I also experienced similar `timeouts` scenarios with a playwright test I wrote recently in another project, It was my first/second time writing tests.

It turned out that the `timeouts` were not the reasons why the tests was failing, it was some `locators` and `assertions` that needed correcting. I have noticed similar pattern here with the tests too, permit me to try some things in the failing portion of the test.

### My Intentions

- Add a `wait` property to the browser `goto` function with a `domcontentloaded` value, this to override the default `load` value.

I would want to try this because, I think the `domcontentloaded` only fires when all html elements are fully loaded and parsed. Well put simple that the DOM has completed its work yea. 🤔

eddiejaoude: We shouldn't need to add any waits/sleeps, these are considered bad practice. I think there is a race condition because different tests fail each time, which is why I think there is a promise not completing properly - as these tests passed correctly before 🤔

eddiejaoude:
I get similar test failures locally, I am still trying to figure out why 🤔

babblebey:
> I get similar test failures locally, I am still trying to figure out why 🤔

Its in fact worse on my end, I'm not even able to run `e2e` tests locally, so I have limited variable to help me debug. I get the timeout all the time, hence my reliance is on the PR actions `build/tests`.

I am also still tryna figure out what's causing the `build/tests` failure. 

Has anyone or have you experienced this `build/tests` failure in any other PR recently?

models/middlewares/dbChangesLogger.js
babblebey: I quite understand, I'm fairly new to testing. Intimidated! I just got the chance to turn my intimidation to strength just 2 weeks ago, when I wrote an `e2e` at https://github.com/open-sauced/app/pull/1551, hence my non-good practice thoughts, 

...but it is quite a thing to me for with the error in the stack trace:

```shell
3 failed
    [chromium] › search.spec.js:21:5 › Search works correctly ──────────────────────────────────────
    [chromium] › search.spec.js:35:5 › Search page has random results when no search term used ─────
    [chromium] › search.spec.js:46:5 › Search page shows random results after typing 1 characters ──
```

```shell
1 failed
    [chromium] › discover.spec.js:19:5 › Discover shows Profiles ───────────────────────────────────
```

I'm wondering, has there been any change made to these routes i.e. `/search` and `/discover`?

eddiejaoude:
I have been chatting with the maintainers, Dan mentioned this `You added middleware into it, so there is something the middleware is doing that breaks things.` and this is the error Dan shared

```
ERROR (1474): failed loading profiles
    err: {
      "type": "TypeError",
      "message": "Cannot read properties of undefined (reading 'find')",
      "stack":
          TypeError: Cannot read properties of undefined (reading 'find')
              at getProfiles (webpack-internal:///./pages/api/profiles/index.js:43:77)
              at async fetchRandomUsers (webpack-internal:///./pages/search.js:52:19)
              at async getServerSideProps (webpack-internal:///./pages/search.js:77:37)
    }
```

I have will try to dig into it. 

Btw how come you can't run the tests locally?

eddiejaoude:
I am struggling to fix this issue, maybe for the moment, we don't use the middleware and we manually save the changes in the API when we need it? What do you think?

babblebey:
> Btw how come you can't run the tests locally?

It's weird 😐

______

> I am struggling to fix this issue, maybe for the moment, we don't use the middleware and we manually save the changes in the API when we need it? What do you think?

Hmmm, I think I subscribe to that as a make due for now.

So this means we will be consuming this functionality within the `page/api/account/manage/link/[[...data]]` and `page/api/account/manage/profile` route!? 

Kindly confirm if there's any other place where the functionality will be implemented apart from this!?

eddiejaoude:
Thanks for understanding!

I think all the CRUD on the manage pages
- profile
- links
- milestones
- events
- testimonials
- repos

https://github.com/EddieHubCommunity/BioDrop/tree/main/pages/api/account/manage

babblebey:
> I think all the CRUD on the manage pages

Hi @eddiejaoude, I would've implemented this like so..

- Create a new function that replicates the `dbChangeLogger` function's functionality, but 

    ### The BIG BUT
    The function MAY not be able to compute the document `changesBefore` & `changesAfter` due to the position it seats in the operation (not in the middle, like the middleware), but it will be able to accept the `changesBefore` & `changesAfter` as arguments, then it can compute a value for the `diff` property with these and save to the `changelog` model.
    
    Now this means we will have to do a `findOne` in each endpoint (atleast for the DELETE & PUT methods) to compute value for `changesBefore` and we can use the returned doc from the endpoints main operation as value for `changesAfter`.

- The function will accept 4 atleast for arguments, i.e. `username`, `collection`, and the aforementioned `changesBefore` & `changesAfter`.

- We will then call this function in every endpoint operation as needed.
- This means the work of the function will be limited strictly to...
  - Computing value for the `diff` property
  - Saving changes to the `changelog` model
  - ...and logging changes to the console

### My Reservation

Having a function that does this operation seating in the middle of the system still looks best for this, because I don't really subscribe to the repetitive call on the function and outside computation of the `changesBefore` & `changesAfter` in each endpoint.

### Way forward

I will go ahead and implement this though and see if it even works!

But I kinda recalled a certain functionality in my experience working with `firebase`, it's known as `snapshots` firebase holds this by default keeping the most previous state of each document and after a change. It got me to research, seeking similar functionality in `mongodb`, hence I found the MongoDB `Change Stream Pre- and Post-Images for Documents`...

https://www.mongodb.com/docs/v6.0/changeStreams/#change-streams-with-document-pre--and-post-images

...introduced in mongoDB 6.0, they are said to be...

> Unlike [middleware](https://mongoosejs.com/docs/middleware.html), change streams are a MongoDB server construct, which means they pick up changes from anywhere.
>
> Quoted from [Change Stream - Mongoose](https://mongoosejs.com/docs/change-streams.html)

...sounds like the dream, but in order to enable them.. we'd need to enable it at database level, this will mean that you'd need to enable something in the production database at merge. Hence I thought to ask to look at it first.

While you take a look, I will work on our make due functionality anyways like I said!

eddiejaoude:
Thank you for the great explanation! I wonder if we could step in this direction incrementally, to allow us to get value out quicker and also de-risk the changes

What about a function that only saves the new changes? We can add further functionality later

babblebey:
> Thank you for the great explanation! I wonder if we could step in this direction incrementally, to allow us to get value out quicker and also de-risk the changes
> 
> What about a function that only saves the new changes? We can add further functionality later

So I have implemented the handy function `logChange` with the specification I initially stated..

```js
export default async function logChange({ username, collection, changesBefore, changesAfter }) {
  let diff;

  const collectionName = collection;
  const docId = changesAfter._id || changesBefore._id;

  if (changesBefore && changesAfter) {
    diff = getChangesDiff(changesBefore, changesAfter);
  } else {
    diff = changesAfter;
  }

  const change = {
    username,
    docId,
    collectionName,
    changesBefore,
    changesAfter,
    diff,
  };

  if (changesBefore !== changesAfter) {
    await Changelog.create({ ...change });
    logger.info(change);
  } else {
    logger.info(`No changes occured in: ${collection}`);
  }
}

function getChangesDiff(chngB, chngA) {
  ...
}
```

and it work exactly as expected. I have only implemented it in Enpoint with "PUT" and "POST" method, because I thought we need add a `Changelog` for deleted (DELETE method) documents.

Here's the PUT Method implementation style in `updateLinkApi` endpoint

```js
export async function updateLinkApi(username, id, data) {
  ...

  // get Link before Update action
  let beforeUpdate = await getLinkApi(username, id);
  
  // ... update action occurs

  // Add to Changelog
  logChange({
    username, 
    collection: "links",
    changesBefore: beforeUpdate,
    changesAfter: getLink
  });

  return JSON.parse(JSON.stringify(getLink));
}
```

...and POST Method implementation style in `addLinkApi` endpoint

```js
export async function addLinkApi(username, data) {
  ...

  getLink = await getLinkApi(username, getLink[0]._id);

  // Add to Changelog
  logChange({
    username, 
    collection: "links", 
    changesBefore: null, 
    changesAfter: getLink
  });

  return JSON.parse(JSON.stringify(getLink));
}
```

I have done this implementation for the following account manage api routes

- profile
- links
- milestones
- events
- repos

I have left out the `testimonial` because I think that updatingTestimonialOrder and its Pinned status isn't much worthy of an addition to the `changelog` as much as links click incrementation doesn't.

eddiejaoude:
> I have left out the testimonial because I think that updatingTestimonialOrder and its Pinned status isn't much worthy of an addition to the changelog as much as links click incrementation doesn't.

Good idea 👍  Thank you for the great work, I will take a look over the next few days - I am at a conference so might be a bit slower than usual

babblebey:
> I will take a look over the next few days - I am at a conference so might be a bit slower than usual

Well noted Eddie!

I hope you have them good sessions.

models/middlewares/dbChangesLogger.js
```diff
@@ -0,0 +1,74 @@
...
+  const diff = {};
+
+  for (const key in chngA) {
+    if (JSON.stringify(chngB[key]) !== JSON.stringify(chngA[key])) {
+      diff[key] = chngA[key];
+    }
+  }
+
+  return diff;
+}
```
eddiejaoude, line 74: With the new approach is this file still needed?

models/Changelog.js
```diff
@@ -0,0 +1,15 @@
+import mongoose from "mongoose";
+ 
+const ChangelogSchema = new mongoose.Schema(
+  {
+    username: String,
```
eddiejaoude, line 5: Sorry to make another change, could the `user._id` be added to the field `userId` in this model please

models/middlewares/dbChangesLogger.js
babblebey: No longer needed yea! I'll remove!

models/Changelog.js
babblebey: Oh, that's true.

I'll appropriate.

eddiejaoude: thank you 👍

babblebey: Hi @eddiejaoude, I hope you've been well.

I couldn't figure out how to get the `userId`, I have attributed this to my limited knowledge of what the `id` looks like. But it is important to state that I found a certain `id` in the `session` object...

```shell
user: {
  "name": "Olabode Lawal-Shittabey",
  "email": "babblebey@gmail.com",
  "image": "https://avatars.githubusercontent.com/u/25631971?v=4",
  "id": 2563**** // <- THAT ID!?
}
expires: "2023-10-11T17:56:10.427Z"
accessToken: "g******u49Hv5n***************DR******i"
username: "babblebey"
```

...but its a `8 character` id, I obviously think that the `userId` was going to match the character length of the `mongoose.Schema.Types.ObjectId` type as seen in the `users` and `profiles` collection in the `db`. 

Kindly confirm, if this is the `id` in `session.user.id` that you had in mind that we use or is there a part I should be looking at for the id?

babblebey:
Hi @eddiejaoude 

Following some commit and update I performed on the branch, I have been having challenge spinning up the local dev server, I get the error below

```shell
./config/stripe.js:1:0
Module not found: Can't resolve 'stripe'
Did you mean './stripe'?
Requests that should resolve in the current directory need to start with './'.
Requests that start with a name are treated as module requests and resolve within module directories (node_modules, C:\Users\Bode\Documents\Projects\Open Source\LinkFree).
If changing the source code is not an option there is also a resolve options called 'preferRelative' which tries to resolve these kind of requests in the current directory too.
> 1 | import Stripe from "stripe";
  2 | 
  3 | import { serverEnv } from "@config/schemas/serverSchema";
  4 | 

https://nextjs.org/docs/messages/module-not-found

Import trace for requested module:
./pages/api/auth/[...nextauth].js
```

image

I noted that there's been some new package added for the premium BioDrop, `stripe` been one I noted and for this reason I have ran `npm install` in attempt to ensure I have all dependencies files up to date, But the error persists even after restarting the server multiple time.

Kindly confirm, is there an extra step in configuration I should be taking to allow be set-up properly?

eddiejaoude:
@babblebey I think your `.env` file is out of date. I think it is best to have a look at the `.env.example` file and add the missing params to your `.env` - the new stripe values can be left empty but they need to exist in your `.env` file

babblebey:
> I suggest update your branch from main and I think your .env file is out of date. I think it is best to have a look at the .env.example file and add the missing params to your .env - the new stripe values can be left empty but they need to exist in your .env file

Hahaha @eddiejaoude, I did do this! I added the Stripe Variable to my `.env` with empty value but I didn't restart my dev server to even see whether it was gonna work, before I left the screen to drop the comment, lazy me... I knew to do that when I got the error in the stack trace pointing to the server `env` schema, I notice the `zod` definition for the Stripe variable's `.optional()`.

Well I'll do and proceed! 

________

> During testing, when I update my profile name, I get the following error

Yea, that was the error I noticed and I wanted to fix before I came across the Stripe issue that got me to stop working. But I know exactly what to work on... 

The error is from within the `getDiff`, the document it's getting inside the `changesBefore/After` arguments, has present within them some mongo hardened stuffs that the `JSON.stringify` cannot work on.

My fix will be to convert the `changesBefore/After` document argument to plain object with the `toObject()` mongoose method, that way the `getDiff` receives a plain flattened object with no mongo stuffs for the function to work with.

eddiejaoude:
Ok cool, let me know how it goes. Thanks again for working through the challenges of this difficult issue

babblebey:
Hi @eddiejaoude,

This should hopefully conclude the process.

I ended up not using the `toObject()` function, as I uncovered that my data for the `changesAfter` in the Update endpoint were somewhat late, not getting the latest doc after change, I sorted that and saw that I was in fact getting the right document format after implementing a call that compute for the new doc, so no need for the `toObject()` function.

### Discovered Edge Case

I have also discovered a situation where `changelog` gets registered for a document anyway even when there isn't any intentionally inputted new value detected. This is because when a user saves the same info as is on a form without change, the `updatedAt` property on the document changes and that is recorded as a change, registered by the `getChangesDiff` function hence trigger an insert to the `changelogs` collection...

#### The Fix ✅ 

I fixed this by removing the `updatedAt` property from the returned `diff` object from the `getChangesDiff` function hence ensuring an empty object is returned when the `updatedAt` property is the only registered difference.

```js
function getChangesDiff(chngB, chngA) {
  const diff = {};

  for (const key in chngA) {
    if (JSON.stringify(chngB[key]) !== JSON.stringify(chngA[key])) {
      diff[key] = chngA[key];
    }
  }

  const { updatedAt, ...cleanedDiff } = diff;     // <- RIGHT HERE - Destructure out the `updatedAt` property if present

  return cleanedDiff;     // <- and returning the other changes, could be no change at all leading to empty object -> {}
}
```

I then proceed to append an extra check to the `changelog` save condition which confirm whether the return returned `diff` object from the `getChangesDiff` function is empty, hence taking the necessary step to either record a new `changelog` or ignore...

```js
export default async function logChange({ userId, collection, changesBefore, changesAfter }) {
  let diff;

  .....

  if (changesBefore !== changesAfter && JSON.stringify(diff) !== "{}") {    // <- Check whether `diff` is not an empty object 
    await Changelog.create({ ...change });
    logger.info(change);
  } else {
    logger.info(`No changes occured in: ${collection}`);   // <- Logs message when `diff` is an empty object 
  }
}
```

BOOM 💥

eddiejaoude:
Thank you so much! It looks like a great idea 👍 I will test today and hopefully we can merge 🎉 Your efforts on this challenge have been amazing 💪

models/middlewares/logChange.js
```diff
@@ -0,0 +1,45 @@
...
+    docId,
+    collectionName,
+    changesBefore,
+    changesAfter,
+    diff,
+  };
+
+  if (changesBefore !== changesAfter && JSON.stringify(diff) !== "{}") {
+    await Changelog.create({ ...change });
+    logger.info(change);
```
eddiejaoude, line 27: I think this logs too much info, it creates pages each time. Maybe something like this...

```suggestion
    logger.info(`Changelog added for: "${collection}"`);
```

models/middlewares/logChange.js
```diff
@@ -0,0 +1,45 @@
...
+  } else {
+    logger.info(`No changes occured in: ${collection}`);
```
eddiejaoude, line 29: What about? 

```suggestion
    logger.info(`No changelog for: "${collection}" by user id ${userId}`);
```

models/middlewares/logChange.js
```diff
@@ -0,0 +1,45 @@
...
+  for (const key in chngA) {
+    if (JSON.stringify(chngB[key]) !== JSON.stringify(chngA[key])) {
+      diff[key] = chngA[key];
+    }
+  }
+
+  const { updatedAt, ...cleanedDiff } = diff;
+
+  return cleanedDiff;
+}
```
eddiejaoude, line 45:
```suggestion
}

```

models/middlewares/logChange.js
babblebey: Makes sense 😁

pages/api/account/manage/event/[[...data]].js
```diff
@@ -21,6 +25,15 @@ export default async function handler(req, res) {
...
   }
 
+  try {
+    const account = await getAccountByProviderAccountId(session.user.id);
+    if (account) {
+      userId = account._id
+    }
+  } catch (error) {
+    logger.error(error, `Changelog - failed to get account for username: ${username}`);
+  }
```
Comment (eddiejaoude) [inline pages/api/account/manage/event/[[...data]].js:35]: Not all requests come via http, they can also go direct to the functions below. I think this needs moving into each function?

pages/api/account/manage/event/[[...data]].js
```diff
@@ -115,6 +130,15 @@ export async function updateEventApi(username, id, updateEvent) {
     return { error };
   }
 
+  
+  // Add to Changelog
+  logChange({
+    userId, 
+    collection: "events", 
+    changesBefore: beforeUpdate, 
+    changesAfter: await getEventApi(username, id)
```
Comment (eddiejaoude) [inline pages/api/account/manage/event/[[...data]].js:139]: Can 1 more field be added to the model, for the string `CREATE`, `DELETE`, `UPDATE`

eddiejaoude (changes requested):
Looks and works great 👍 

Sorry I thought of 2 more things, I think they are the last 2 things

pages/api/account/manage/event/[[...data]].js
Comment (babblebey) [reply pages/api/account/manage/event/[[...data]].js:139]: > Can 1 more field be added to the model, for the string `CREATE`, `DELETE`, `UPDATE` 

Hmm, Okay!

So the new field would be meant to track the CRUD Operation Type!?

`operation` good for a name!?

________

> - then can we add the `logChange` to the create and delete functions also?

We already have it integrated in the `create` endpoints for the respective models, do you mean something different when you say "functions"?

Also, I'd intentionally not integrated the `logChange` in the `delete` endpoints, because it's a deletion, is there a reason or a usecase for the deletion operation changelog!? I'd like a peek into your thoughts, I'm curious 🤓

pages/api/account/manage/event/[[...data]].js
Comment (babblebey) [reply pages/api/account/manage/event/[[...data]].js:35]: D'you mean I should run the `getAccountByProviderUserId` to populate the `userId` separately within each endpoint function!? 🤔

pages/api/account/manage/event/[[...data]].js
Comment (eddiejaoude) [reply pages/api/account/manage/event/[[...data]].js:139]: > operation good for a name!?

sounds good 👍 

---

> We already have it integrated in the create endpoints for the respective models, do you mean something different when you say "functions"?

For example In the events API file, the logchange is not on the function that does the create and delete, I think if we displayed a change log to admins or the user it would be super helpful

Simple example we might display

| Section | Date | Before | After | Action |
| :--- | :--- | :--- | :--- | :--- |
| Events | 12th September 2023 14:34 | - | `{...}` | Created |
| Events | 13th September 2023 14:34 | `{...}` | `{...}` | Updated |
| Events | 14th September 2023 14:34 | `{...}` | - | Deleted |

> Also, I'd intentionally not integrated the logChange in the delete endpoints, because it's a deletion, is there a reason or a usecase for the deletion operation changelog!? I'd like a peek into your thoughts, I'm curious 🤓

For example when I delete an event, it would be good to know the latest state of that db entry before it is deleted.

pages/api/account/manage/event/[[...data]].js
Comment (eddiejaoude) [reply pages/api/account/manage/event/[[...data]].js:35]: Yes, because some of these functions are called directly within the app - you could add a new function in the file that is called in each function

Comment (babblebey) [reply pages/api/account/manage/event/[[...data]].js:35]: Oh, this means that the `called directly` do not go through the controller/handler function in each endpoint... 🤔 

```js
  ....
}
```

...correct @eddiejaoude?

pages/api/account/manage/event/[[...data]].js
Comment (eddiejaoude) [reply pages/api/account/manage/event/[[...data]].js:139]: > and the logChange integrations missing in the events endpoint function!? 😳 I must've missed that part oh wow

no I don't think you missed it, I think if the developer doesn't use http and calls the function directly with nextjs, then the userId won't get populated

pages/api/account/manage/event/[[...data]].js
Comment (eddiejaoude) [reply pages/api/account/manage/event/[[...data]].js:35]: yes correct 👍  this can happen with nextjs as it is recommend not to go via http (the handler) for server side calls

pages/api/account/manage/event/[[...data]].js
Comment (babblebey) [reply pages/api/account/manage/event/[[...data]].js:139]: Oh, I knew that... I thought you said that the `logChange` function wasn't called/implemented in the `event` endpoint function like we did in the other endpoint

pages/api/account/manage/event/[[...data]].js
Comment (babblebey) [reply pages/api/account/manage/event/[[...data]].js:35]: Oh, that's true!

But, my reservation now would be this! will be able to retrieve the `session` object without the a `context` object!? 🤔 

```js
const session = await getServerSession(req, res, authOptions)
```

Put simple, I will require the `session.user.id` property...

Maybe i cannot use the `getServerSession` but I can use the `getSession` but is that not meant for client-side? Plus I understand if it will be used on the server then...

> NOTE: When calling getSession() server side, you need to pass {req} or context object. 
https://next-auth.js.org/v3/getting-started/client#getsession

Any ideas!?

Comment (eddiejaoude) [reply pages/api/account/manage/event/[[...data]].js:35]: We had this challenge somewhere else and came up with a solution, let me try to find it in the project

Comment (babblebey) [reply pages/api/account/manage/event/[[...data]].js:35]: Oh great resource!

I'll look to implement with this style.

Comment (eddiejaoude) [reply pages/api/account/manage/event/[[...data]].js:35]: Thank you 👍  also let us know if this can be improved

eddiejaoude:
I just had a test/play, it works really great! 👍  thank you

I think it is just the `userId` remaining to be done. If you need any help on this let me know.

I did notice, that `collection` is probably better renamed to `model`, as most of these changes are to the collection `profile` - otherwise this will become confusing.

Also can the testimonials and premium toggles be added to the changelog? This can come in a later PR though

babblebey:
> I just had a test/play, it works really great! 👍  thank you

😁

> I think it is just the `userId` remaining to be done. If you need any help on this let me know.

Yea, I'm working on that, I've been experimenting with `getSession` function (accepts only `req` as args) which turns out to as not advised to use in SSR from NextAuth v4.***, So I am opting for `getServerSession` (accepts `req`, `res` & `authOptions`) and working on a manner that reduces the number in lines of code to write per endpoints as we try get the `userId` through the `session.user.id`.

> I did notice, that `collection` is probably better renamed to `model`, as most of these changes are to the collection `profile` - otherwise this will become confusing.

Hmmm, maybe!? I struggle to understand the `profile` example though, do you want to give me another example of make this simpler for me 🥲

> Also can the testimonials and premium toggles be added to the changelog? This can come in a later PR though

Hmmm, I seriously struggled to find the area where testimonials had a need for changelog addition in the initial phase, and are testimonials editable after getting added too??

Also, what are the premium toggles!? Sounds very new to me!

eddiejaoude:
Oh yes you are right, the userId is in the session 👍  `session.user.id` - I think using this is find on those pages because the user must be logged in to go to the manage pages - good spot 💯 

Testimonials and Premium features only have toggles, they are all set to "off" to begin with but the user can enable them - it would be good to log when these are toggled on/off

If you need more clarification or want to discuss let me know - we can always continue on EddieHub Discord it might be easier

babblebey:
Great stuff @eddiejaoude,

I now see the user's `accountId` directly as the value `user.id` in `session` object, 

```shell
{
  user: {
    name: 'Olabode Lawal-Shittabey',
    email: 'babblebey@gmail.com',
    image: 'https://avatars.githubusercontent.com/u/25631971?v=4',
    id: '64d6c69f1111111a3934ba94' // <- HERE
  },
  expires: '2023-10-30T12:50:08.590Z',
  accessToken: 'gho_*******u8rF***ucsr********************p***',
  username: 'babblebey',
  accountType: 'free',
  stripeCustomerId: undefined
}
```

I think that's pretty straightforward, instead of having the `session.user.id` value be the `providerId`. It has now indeed shaped better the way the endpoint function will call the `logChanges` function and how other parts of the project will call the endpoint function directly in `SSR` or `GetServerSideProps`. 

The `logChanges` will not look like this...

```js
async function logChange(session, { model, changesBefore, changesAfter }) {
  // Take that user `accountId` -> `64d6c69f1111111a3934ba94` straightforward.
  let userId = session.user.id 

  .....

  const change = {
    userId,
    .....
  };
}
```

and it will be callable from with the endpoint functions like so...

```js
logChange(await getServerSession(context.req, context.res, authOptions), {
  collection: "links",
  changesBefore: beforeUpdate,
  changesAfter: afterUpdate
});
```

this means whenever the function is call directly directly in `SSR` or `GetServerSideProps`, it can simple just take in the `context` object as the first argument or maybe an object that holds `{ req, res }`, like so...

```js
export async function getServerSideProps(context) {
  await updateLinkApi(context, [username], [id], {data})
  ....
}
```

babblebey:
> Testimonials and Premium features only have toggles, they are all set to "off" to begin with but the user can enable them - it would be good to log when these are toggled on/off

I am attempting to experience the toggles, I am getting the below error on the `/account/manage/profile` route...

```shell
Server Error
Error: Error serializing `.PREMIUM_SUPPORT_URL` returned from `getServerSideProps` in "/account/manage/premium".
Reason: `undefined` cannot be serialized as JSON. Please use `null` or omit this value.

This error happened while generating the page. Any console logs will be displayed in the terminal window.
```
image

Well, I am still tryna trace the endpoint to integrate the `logChange` into for the toggles anyway, and I have uncovered the  `updateSettingsApi` function in the `api/account/manage/settings` route, kindly confirm if I am looking at the right function or maybe you could point me in the right direction.

eddiejaoude:
> I now see the user's accountId directly as the value user.id in session object,

note that `accountId`, `userId` and `profileId` are all different, we should use `userId` (user.id) here 

> So, I am attempting to experience the toggles, but I am getting the below error 

I think the error is because you are missing environment variables in your `.env` file, please compare with `.env.example` file - this value can be empty but the property needs to exist in the `.env`

babblebey:
> note that `accountId`, `userId` and `profileId` are all different, we should use `userId` (user.id) here.

Yea, for sure I knew that... I was just talking about the exact value in the `session.user.id` as it differs from the one that I thought used to be there.

> I think the error is because you are missing environment variables in your `.env` file, please compare with `.env.example` file - this value can be empty but the property needs to exist in the `.env`

`PREMIUM_SUPPORT_URL`!? Okay, I'll appropriate that.

eddiejaoude:
> The Implementation works as expected! All tests I've done passed! You should give it a whirl too 😁

Great work! I will take a look shortly.

> I was just talking about the exact value in the session.user.id as it differs from the one that I thought used to be there.

Here is me logging out the `token.sub` or `session.user.id` and it matches my `user._id` in the database

Screenshot 2023-09-30 at 23 29 36

babblebey:
> Here is me logging out the `token.sub` or `session.user.id` and it matches my `user._id` in the database

Yea, exactly the `id` I was taking about. 

That was thesame value I'd tried to get using the `getAccountByProviderId` function with the `session.user.id` as args... At that point, the value of the `session.user.id` was the GitHub providerId instead.

babblebey:
Following the integration into the `Premium Setting Toggle`, I discovered a certain edge case.

### Edge Case

A Premium User is most certainly gonna want to hide both the `Footer` and `Header`, now In order to perform this operation, the toggle on each setting makes separate `http-requests` to the endpoint, making it at least of 2 requests to perform a one way operation (i.e. false-to-true) on both toggles. This in turn also records at least 2 `changelogs`, which IMO is somewhat unnecessary.

image

### Fix Proposal

I think we can reduce the number of `http-requests` and recorded `changelog` to just 1. This will be the case if we implement a `Save Change` button which only makes the `http-request` and trigger a `changelog` when the user assumes satisfaction with their toggle preferences, and clicks on the `Save Changes` button to confirm their selection/preferences, then the state of the toggles (`true` or `false` on both `Footer` and `Header`) can be sent together in 1 request. This change might be insignificant but I think it will reduce the number of `changelog` to record for this Premium toggle and come'on we are making atleast 1 less `http-request` ahahaha 😃 

image

We can just just have the button there in `disabled` state by default or not even present at all to start with, then as soon as any of the toggle state changes, we can `enable` the button to give a visual que to user notifying them the need to confirm their toggle selection/preferences.

What'd you think @eddiejaoude???

eddiejaoude:
Good suggestion, but I think we can leave it for now, as I am keen to get this PR merged as soon as possible

Thank you for making the changes, I can't wait to have a play

babblebey:
Great stuff 😃

I can't wait either hahaha 😃

So the last leg would be to figure out every instance where the endpoints was indeed called via `non-http` means in SSR.

But judging by the fact that we experience no error at build time, I will conclude that there isn't any of these instances yet.

Give this a whirl and let's see what we got! 😁

### Quick Question

Are you tryna work on some administrator dashboard where this model will be implemented to track changes!? 🤓

Or better still, what's the use of this!? I'm just curious

pages/api/account/manage/event/[[...data]].js
```diff
@@ -114,14 +119,23 @@ export async function updateEventApi(username, id, updateEvent) {
     log.error(e, error);
     return { error };
   }
+  
+  // Add to Changelog
+  logChange(await getServerSession(context.req, context.res, authOptions), {
+    model: "Event", 
+    changesBefore: beforeUpdate,
```
Comment (eddiejaoude) [inline pages/api/account/manage/event/[[...data]].js:126]: `beforeUpdate` is not used anywhere else, so I suggest it can be moved from line 65 to here similar to line 127 with

```suggestion
    changesBefore: await getEventApi(username, id);
```

pages/api/account/manage/event/[[...data]].js
```diff
@@ -145,10 +159,17 @@ export async function deleteEventApi(username, id) {
     return { error };
   }
 
+  // Add to Changelog
+  logChange(await getServerSession(context.req, context.res, authOptions), {
+    model: "Event", 
+    changesBefore: beforeDelete,
```
Comment (eddiejaoude) [inline pages/api/account/manage/event/[[...data]].js:165]: Same here?

pages/api/account/manage/event/[[...data]].js
```diff
@@ -75,10 +78,12 @@ export async function getEventApi(username, id) {
   return JSON.parse(JSON.stringify(getEvent[0]));
 }
 
-export async function updateEventApi(username, id, updateEvent) {
+export async function updateEventApi(context, username, id, updateEvent) {
   await connectMongo();
   const log = logger.child({ username });
 
+  const beforeUpdate = await getEventApi(username, id);
```
Comment (eddiejaoude) [inline pages/api/account/manage/event/[[...data]].js:85]: I think this can be moved to line 126?

eddiejaoude (approved):
Works really well 🎉  thank you! 🔥

eddiejaoude:
I noticed that (I think) testimonial is missing?

pages/api/account/manage/event/[[...data]].js
Comment (eddiejaoude) [reply pages/api/account/manage/event/[[...data]].js:85]: I think these little improvements can come in another PR

babblebey:
> I noticed that (I think) testimonial is missing? This an come in another PR

I saw testimonials wasn't controlled via Form Input, testimonials were linked to creating an Issue on the repository.

Or is that somewhere I should be looking that I didn't look at?? 🤔

eddiejaoude:
I mean for the manage testimonials (not create), they are toggled on/off - it would be good to record those changes to people's profiles

Could you raise an issue? You don't have to work on it, someone else can use your examples

PR #8854, opened 2023-08-29
fix: glow animation in Links section

### LINKED ISSUE #8796: [BUG] Glow animation in Links section is not working in Dark mode
### Description

Glow animation in Links section is not working in Dark mode

### Screenshots

image

### Additional information

Animation works fine in Light mode

zaid-ahmad:
Closes #8796

Glow animation in Links section now works in Dark mode.

msedge_JzF6ExPdpE


eddiejaoude:
Hi @zaid-ahmad did you mean to close your PR?

zaid-ahmad:
I created the PR before getting assigned. So i closed it.

SaraJaoude:
@zaid-ahmad please see my comments here: https://github.com/EddieHubCommunity/BioDrop/issues/8796#issuecomment-1698677964

SaraJaoude:
@zaid-ahmad it would be helpful if you could please add the screen recording you mentioned [here](https://github.com/EddieHubCommunity/BioDrop/issues/8796#issuecomment-1698039462) to this PR.

zaid-ahmad:
> @zaid-ahmad it would be helpful if you could please add the screen recording you mentioned [here](https://github.com/EddieHubCommunity/BioDrop/issues/8796#issuecomment-1698039462) to this PR.

I've added it in the comments.

components/layouts/MultiLayout.js
```diff
@@ -8,7 +8,7 @@ export default function MultiLayout({ children }) {
       <SkipLink />
       <div className="flex flex-col min-h-screen">
         <Navbar />
-        <main id="main" className="flex-1 dark:bg-dark">
+        <main id="main" className="flex-1 dark:bg-dark z-10">
```
eddiejaoude, line 11: Not sure why this change is so far away from the area that it is trying to fix? It could have a bad effect on other places in the app

eddiejaoude (changes requested):
Thank you for the contribution, I have left an inline comment

components/layouts/MultiLayout.js
zaid-ahmad: The reason this was happening was because the z-index of the glow animation is set to -99999. As a result, when a background is set in the dark mode, it goes beneath it. So the easiest way to solve this problem was to set a relative z-index so it is visible.
Now don't worry, I've tested the change for every single page on the app and it works perfectly fine 😄

zaid-ahmad: I had two possible solutions to the problem. The first was to make changes to the component's file (UserLink.js). This would add a lot of code, which could make it difficult to make changes in the future. I avoided this, as it would be more maintainable to keep the code as simple as possible.

The second solution was to change the z-index of the glow animation. However, I tried this and it did not work. I believe this is because the glow animation happens inside the div, and changing the z-index of the glow animation causes it to cover the entire input. This would also cover the logo, which would prevent the glow-icon feature from working.

This is why I went with the third approach of changing the z-index of the parent main itself. Again, I have checked it on all pages, it works perfectly fine.

zaid-ahmad: @eddiejaoude the PR is not yet merged, should I find another way of doing it? 😅

eddiejaoude: yes please 👍 these current changes don't feel quite right

eddiejaoude: Thank you for the detailed explanation 👍 hmm maybe that is a better idea to only apply the z-index in dark mode

zaid-ahmad: I have made the changes to the PR. Hope it gets merged soon! (excited)

eddiejaoude (approved):
Thank you 👍

PR #8857, opened 2023-08-30
Gold star added

JakeDamon:


sital002 (commented):
This is just a suggestion:
Instead of adding a star we could a crown icon similar to this with gold color.
image


Before Changes:
image


After Changes:
image
And on more thing if we are displaying the gold star just after the name anyone can add a emoji similar to star that will cause problem instead if it's a premium user then we should have a tooltip that will show a message like Premium badge on hover on the icon.
image

Here you could see that if I just add emoji to the name field it will create a problem.

components/user/UserProfile.js
```diff
@@ -48,9 +49,11 @@ function UserProfile({ BASE_URL, data }) {
             className="rounded-full object-contain"
           />
         </Badge>
-
         <div className="flex flex-col self-center gap-3">
-          <h1 className="text-3xl font-bold">{data.name}</h1>
+          <h1 
+            className="flex text-3xl font-bold gap-1">{data.name} 
+            <ImStarFull color="gold" />
```
eddiejaoude, line 55: The tests are failing, not sure why they are asking for a title in the icon 🤔 

```
"description": "Ensures each HTML document contains a non-empty <title> element",
```

JakeDamon:
I will add a modal so when clicked, it will display a Premium Account description. Much like X's verified account blue check.

eddiejaoude (changes requested):
Thanks for the contribution Jake 👍 

There seems to be many other changes in your pull request, I think it is because you are using the same branch `premium-strip` that you are merging into, so it is difficult to keep it up to date? Can you fix your PR so it only has your changes in it, thanks

JakeDamon:
Closing to reopen to get latest changes.

PR #8896, opened 2023-09-01
data: fix testimonials that i broke because i had changed my username…

xkrishguptaa:
Okay I got some explanation to do, probably, which I'm not going to do, but here is the fix

it renames my second last username on link free to my current ones which I inconveniently forgot to update the last time in the pr

fixed it @loftwah


loftwah:
image

Cheers Krish 🎉🎉

xkrishguptaa:
no bullying lmao dean

kumarsonsoff3 (commented):
Thanks for renaming these all, Krish! 👍 

Could you please justify why the files related to `web3community.json` are being deleted!?

xkrishguptaa:
web3community.json is being deleted because I made it back when the org was web3community now it is WebXDAO, I am not renaming it because organization profiles are not available in BioDrop rn, so I though removing might be the best idea.

kumarsonsoff3 (approved):
Okay, Thanks! LGTM 👍

PR #8953, opened 2023-09-05
fix:fixed the unable to save a link without specifying an icon

### LINKED ISSUE #8799: [BUG] Unable to save a link without specifying an icon
### Description

Today I was setting up my BioDrop profile. I wanted to add my website as a link. I didn't find an icon that looked right, but I did notice that the preview display of the link showed a generic "World Wide Web" globe when no icon was specified. (See screenshot). So I thought it would be fine. However, I discovered that when I clicked "save," nothing happened. After some trial and error, I found that I had to specify an icon in order to save a link. Since I didn't find a link that looked right, I just went with a Font Awesome Git Alt icon ("FaGitAlt") for now. However, I'd like to be able to submit a link without specifying an icon, and have it default to the "World Wide Web" globe shown in the preview in the screenshot below.

Update: While writing this issue, I discovered that the "World Wide Web" globe icon I wanted is actually available as an icon with the name "FaGlobe." So in some sense that fixes my issue. But I would like it better if I could skip selecting an icon and be assigned the globe. If that's not possible or desired, then I would at least want an error message that explains why I can't save the link (because of there being no icon specified).

Also, I would like to work on this issue as a developer.

### Screenshots

Screen Shot 2023-08-26 at 11 26 37 AM

### Additional information

_No response_

yashnerkar:
## Fixes Issue

Closes #8799

## Changes proposed

I have set the default icon while sending a request to the server so that in case the user has not chosen an icon, the default icon will be `fa-globe`. The reason I didn't directly set it in the state is that other fields are empty, and having only one field set as the icon does not look good. Additionally, we have already set the icon while posting the request to the server.


pages/account/manage/link/[[...data]].js
```diff
@@ -75,7 +75,16 @@ export default function ManageLink({ BASE_URL, username, link }) {
...
 
     let method = "POST";
-    let putLink = { group, name, url, icon, isEnabled, isPinned, animation };
+    let selectedIcon = icon !== "" ? icon : "fa-globe";
+    let putLink = {
+      group,
+      name,
+      url,
+      icon: selectedIcon,
+      isEnabled,
```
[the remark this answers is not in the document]
Comment (yashnerkar) [reply pages/account/manage/link/[[...data]].js:84]: Yeah we can do that, but by creating the variable makes code more readable for other developers also if want to add any logic here, we can add that without cluttering the object.

pages/account/manage/link/[[...data]].js
```diff
@@ -75,7 +75,16 @@ export default function ManageLink({ BASE_URL, username, link }) {
     e.preventDefault();
 
     let method = "POST";
-    let putLink = { group, name, url, icon, isEnabled, isPinned, animation };
+    let selectedIcon = icon !== "" ? icon : "fa-globe";
```
Comment (eddiejaoude) [inline pages/account/manage/link/[[...data]].js:78]: This should be `FaGlobe` because we use React icons 

```suggestion
    let selectedIcon = icon !== "" ? icon : "FaGlobe";
```

eddiejaoude (changes requested):
Thank you 👍 I left an inline comment

pages/account/manage/link/[[...data]].js
Comment (yashnerkar) [reply pages/account/manage/link/[[...data]].js:78]: okay! will change it

yashnerkar:
changed the icon name to FaGlobe

eddiejaoude (approved):
Thank you 👍

PR #9085, opened 2023-09-16
[BUG] Truncated long urls in /account/statistics page

### LINKED ISSUE #8952: [BUG] In the statistics, design breaks with long links
### Description

If you have some long link in your link list, on the panel that shows click number on each link, the link line doesn't break so the card containing the links get larger and larger.

### Screenshots

Screenshot 2023-09-04 164133

Yes

### Additional information

To fix this it would be enough to wrap the content or to add an overflow handling. Since those aren't clickable link but just informative, I feel like the wrap solution would work better, what do you think?

theunhackable:
## Fixes Issue
Closes #8952

## Changes proposed
- Fixed unresponsiveness and overflow of long links
- Added expanding and collapsing feature

## Change requested
- requested to truncate the link instead of the complex solution.

## Final changes

- added truncate class from tailwindCSS as per the recomendations and solve the overflow issue by adding max-width.
- 

## Screenshots
Truncated link 👇:
image

## Note to reviewers

Responsiviness if fixed. I have not seen any issue raised on this so i fixed if without raising an issue. If there are any open issues that I am unable to see about the responsiveness of the link please close it too.

### DIFF
--- a/pages/account/statistics.js
+++ b/pages/account/statistics.js
@@ -17,36 +17,11 @@ import { abbreviateNumber } from "@services/utils/abbreviateNumbers";
 import Navigation from "@components/account/manage/Navigation";
 import UserMini from "@components/user/UserMini";
 import { PROJECT_NAME } from "@constants/index";
-import { useState } from "react";
 
 const DynamicChart = dynamic(
   () => import("../../components/statistics/StatsChart"),
   { ssr: false }
 );
-
-const ExpandingUrl = ({url}) => {
-  const [isExpanded, setIsExpanded] = useState(false);
-
-  return (
-    url.length > 40?(
-    <>
-      {url.substring(0, 40) + (isExpanded === false ? (" ...") : (""))}
-        
-      {isExpanded && (
-          url.substring(40, url.length)
-      )}
-
-      <div onClick={() => setIsExpanded(prev => !prev)} className="hover:cursor-pointer text-primary-low-high">
-        {isExpanded === false ? ("click to expand") : ("click to collapse")}
-      </div>
-  
-    </>
-    ):(
-      url
-    )
-  )
-    
-}
 
 export async function getServerSideProps(context) {
   const { req, res } = context;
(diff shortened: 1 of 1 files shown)


eddiejaoude (changes requested):
Thank you for the contribution but I think we should try to achieve this with css as this looks over complicated

theunhackable:
One thing I could do is just to limit the link to certain fixed characters. Can you please eloborate your way of using CSS in this? Any reference would help

eddiejaoude:
Here are the docs to Tailwind, we can benefit from the tool we use without needing to code it all rom scratch https://tailwindcss.com/docs/text-overflow

pages/account/statistics.js
```diff
@@ -272,7 +272,7 @@ export default function Statistics({ data, profile, progress, BASE_URL }) {
             {data.links &&
               data.links.individual.map((link) => (
                 <tr key={link.url}>
-                  <td className="md:whitespace-nowrap py-4 pl-4 pr-3 text-sm font-medium text-primary-high dark:text-primary-low sm:pl-6">
+                  <td className=" truncate break-words max-sm:max-w-[200px] max-w-lg bg-dark py-4 pl-4 pr-3 text-sm font-medium text-primary-high dark:text-primary-low">
```
eddiejaoude, line 275: Is the px width needed? Seems fragile, it should automatically truncate on overflow

theunhackable: it doesn't interfere with any modern devices as 99.99999999% devices are above width 320px. I'm just making sure that there is atleast some breathing area is given to the link inorder to make it readable. I don't think it is a problem.

eddiejaoude: px measurement is not great, if not needed most of the time best to remove it - we don't need to over optimise

theunhackable: image
 
For some strange reason the table is overflowing after removing the breakpoint.

eddiejaoude:
Thank you for your contributing but I will close this as there has been no activity for 4+ weeks and there are conflicts

Issue #9102, opened 2023-09-18
[UI BUG] The hero section buttons layout is squeezed on mobile devices.


hi-Kartik2004:
### Description

The hero section buttons layout is squeezed on mobile devices.
PFB the screenshots for better understanding and details.

### The hero section (Now)
![image](https://github.com/EddieHubCommunity/BioDrop/assets/111000515/e4f29b89-e675-40b5-939a-28a3dab1b0aa)

![image](https://github.com/EddieHubCommunity/BioDrop/assets/111000515/4157918e-1c8a-4e00-b430-48791bc4d540)

### Hero section (How it should actually look)
![image](https://github.com/EddieHubCommunity/BioDrop/assets/111000515/0da0e5b2-4b75-4d40-b303-8325e0e73edf)

### Do you want to work on this issue?

Yes

### Additional information

Reviewing the issue I found out that there is a conditional margin given to this section, removing which will solve this issue and the hero section button wrapper would look like the desired screenshot above.

You could check this out with your browser dev tools!

eddiejaoude:
Thank you, I have assigned it to you 👍

PR #9159, opened 2023-09-25
Fixed hero-buttons layout on Mobile devices.

hi-Kartik2004:

- [] My change requires changes to the documentation.

#### How it looks now:
image

## Note to reviewers
You could refer the issue #9102 for more information.

### DIFF
--- a/pages/index.js
+++ b/pages/index.js
@@ -232,7 +232,7 @@ export default function Home({
                 <FaDollarSign /> Free Tier Forever
               </div>
             </div>
-            <div className="mt-10 flex flex-col sm:flex-row items-center justify-center gap-6 mx-24">
+            <div className="mt-10 flex flex-col sm:flex-row items-center justify-center gap-6 sm:mx-24">
               <Button href="/docs" primary={true}>
                 Get started
               </Button>


pages/index.js
```diff
@@ -232,7 +232,7 @@ export default function Home({
                 <FaDollarSign /> Free Tier Forever
               </div>
             </div>
-            <div className="mt-10 flex flex-col sm:flex-row items-center justify-center gap-6 mx-24">
+            <div className="mt-10 flex flex-col sm:flex-row items-center justify-center gap-6 sm:mx-24">
```
Dun-sin, line 235: how about making it more dynamic, like this instead of a static value
```suggestion
            <div className="mt-10 flex flex-col sm:flex-row items-center justify-center gap-6 min-w-[300px] max-w-[700px] w-[50%]">
```

eddiejaoude: Yes I agree, it is never good to add px

eddiejaoude (changes requested):
Thank you for your contribution. We added, inline comments

pages/index.js
hi-Kartik2004: Hi, the suggested changes leads to the below output:
image

The issue is that `min-width: 300px` will cause the website UI to overflow for lower screen width devices like samsung galaxy fold (Screen width: 280px) etc.

`max-width: 700px` does nothing as the container / wrapper of the button layout already has the max-width property.

`width: 50%` will cause useability issues on lower mobile devices.
The current code is fully responsive and I don't find any reasons to implement the suggest changes.
Please let me know if I have missed something.

SaraJaoude: I have marked as this as `unresolved` as it is still an ongoing discussion.

eddiejaoude:
Did a force push happen here? It looks like the history has changed

hi-Kartik2004:
> Did a force push happen here? It looks like the history has changed
just updated the branch, I haven't made any code changes, since the I raised this PR.

pages/index.js
eddiejaoude: After reviewing the PR, I think `sm:` does work better here

PR #9248, opened 2023-10-01
fix: resolve mobile keyboard fluctuation during search

### LINKED ISSUE #9208: [BUG] Mobile Keyboard Fluctuation During Search
### Description

In search page when performing a search on my mobile device, the on-screen keyboard initially appears, then unexpectedly disappears and reappears during the search process, causing an inconsistent user experience and making it challenging to input text efficiently.

**Environment**
OS: Android
Browser: Google Chrome

### Screenshots

_No response_

Yes

### Additional information

_No response_

badrivlog:
## Fixes Issue
closes #9208 

### Implemented URL search params search functionality
When a user searches for user profile and submit the search form or clicked on the tag, the URL params will be updated, data will be fetched on the server, and the search page will re-render with the new data.

## Screenshots

https://github.com/EddieHubCommunity/BioDrop/assets/73095526/3b10ef61-9ed1-46fe-b2d3-cad4b486f435


pages/search.js
```diff
@@ -145,13 +145,11 @@ export default function Search({
...
     const timer = setTimeout(() => {
-      router.replace(
-        {
-          pathname: "/search",
-          query: { userSearchParam: inputValue },
-        },
-        undefined,
-        { shallow: true },
+      const newUrl = `/search?userSearchParam=${inputValue}`;
+      window.history.replaceState(
```
eddiejaoude, line 149: Is it not possible with react/next router? https://nextjs.org/docs/pages/building-your-application/routing/linking-and-navigating#shallow-routing

badrivlog: I tried it, but didn't work as expected and currently we're using this method but page gets reloaded causing the mobile keyboard close and open during search

badrivlog: My idea is update the url without page gets reloaded and it works fine in this case

badrivlog: > Is it not possible with react/next router? https://nextjs.org/docs/pages/building-your-application/routing/linking-and-navigating#shallow-routing

According to nextjs docs `shallow-routing` is probably only relevant if our page uses `getStaticProps`, `getServerSideProps` or `getInitialProps` to avoid re-running these data fetching methods again when the url change. But unfortunately `next/router` still re-rendering the page. 

In my research, I came across this https://github.com/vercel/next.js/discussions/18072 discussion thread to solve this issue.

badrivlog:
Hey, any news regarding this PR? Thanks.

badrivlog:
@eddiejaoude any advice or guidance regarding this PR

eddiejaoude:
Sorry I am moving house and country so behind on notifications at the moment

sital002 (commented):
Instead of using ```useEffect``` we can use ```useLayoutEffect``` https://github.com/EddieHubCommunity/BioDrop/pull/9248/files#diff-99c55e6229f7a0d904e59da5da6e77729e62b05852484cf2afff2e13292b50b1R96-R101 but useLayoutEffect but it has some drawbacks here is the link for ```useLayoutEffect``` https://react.dev/reference/react/useLayoutEffect

pages/search.js
```diff
@@ -93,10 +99,8 @@ export default function Search({
       setNotFound(`${username} not found`);
     }
   }, [username]);
-  useEffect(() => {
-    if (searchInputRef.current) {
-      searchInputRef.current.focus();
-    }
+  useLayoutEffect(() => {
+    searchInputRef.current?.focus();
```
sital002, line 103:
```suggestion
    searchInputRef?.current.focus();
```
i think this is the correct way

badrivlog: `searchInputRef` is all ready exist, why do need check for this I'm confused

badrivlog:
Thanks @sital002 for your great suggestion👍 @eddiejaoude can you take a look now

pages/search.js
sital002: we are checking if the current exits or not

eddiejaoude (commented):
Great collaboration 👍  it works well

I notice there is an error in the terminal from these changes though...

```js
Warning: useLayoutEffect does nothing on the server, because its effect cannot be encoded into the server renderer's output format. This will lead to a mismatch between the initial, non-hydrated UI and the intended UI. To avoid this, useLayoutEffect should only be used in components that render exclusively on the client. See https://reactjs.org/link/uselayouteffect-ssr for common fixes.
    at Search (webpack-internal:///./pages/search.js:90:16)
    at main
    at div
    at MultiLayout (webpack-internal:///./components/layouts/MultiLayout.js:17:24)
    at SessionProvider (/Users/ej/Downloads/repos/EddieHub/LinkFree/node_modules/next-auth/react/index.js:454:24)
    at m (/Users/ej/Downloads/repos/EddieHub/LinkFree/node_modules/next-themes/dist/index.js:1:310)
    at exports.ThemeProvider (/Users/ej/Downloads/repos/EddieHub/LinkFree/node_modules/next-themes/dist/index.js:1:3611)
    at MyApp (webpack-internal:///./pages/_app.js:30:18)
    at StyleRegistry (/Users/ej/Downloads/repos/EddieHub/LinkFree/node_modules/styled-jsx/dist/index/index.js:449:36)
    at ek (/Users/ej/Downloads/repos/EddieHub/LinkFree/node_modules/next/dist/compiled/next-server/pages.runtime.dev.js:30:13126)
    at eY (/Users/ej/Downloads/repos/EddieHub/LinkFree/node_modules/next/dist/compiled/next-server/pages.runtime.dev.js:39:1766)
    at eV (/Users/ej/Downloads/repos/EddieHub/LinkFree/node_modules/next/dist/compiled/next-server/pages.runtime.dev.js:39:3110)
    at div
    at e1 (/Users/ej/Downloads/repos/EddieHub/LinkFree/node_modules/next/dist/compiled/next-server/pages.runtime.dev.js:48:761)
```

sital002:
I think we should wrap the input in a form and only update the URL on submit instead on change since we are planning to do a db call for searching users using on change will have unnecessary db calls this will solve the fluctuation error the issue is whenever we update the query params using ```next router``` the default behavior is that it rerenders the whole page this means it is running all the ```useEffect``` every time the input data changes and the ```focus```  is rerunning.

badrivlog:
> I think we should wrap the input in a form and only update the URL on submit instead on change since we are planning to do a db call for searching users using on change will have unnecessary db calls this will solve the fluctuation error the issue is whenever we update the query params using ```next router``` the default behavior is that it rerenders the whole page this means it is running all the ```useEffect``` every time the input data changes and the ```focus```  is rerunning.  

Thank you for your co-operation. That's a great idea👍

I would like to hear other maintainers thoughts on it since this will change searching behavior

eddiejaoude:
> I think we should wrap the input in a form and only update the URL on submit instead on change since we are planning to do a db call for searching users using on change will have unnecessary db calls this will solve the fluctuation error the issue is whenever we update the query params using `next router` the default behavior is that it rerenders the whole page this means it is running all the `useEffect` every time the input data changes and the `focus` is rerunning.

I think this is a good idea also 👍

pages/search.js
```diff
@@ -95,29 +97,31 @@ export default function Search({
...
+        e.preventDefault();
+        searchInputRef.current?.focus();
+      }
+    };
+    document.addEventListener("keydown", onKeyDownHandler);
+
+    return () => {
+      document.removeEventListener("keydown", onKeyDownHandler);
+    };
+  }, []);
```
badrivlog, line 111: not sure why this cryptic changes?

pages/search.js
```diff
@@ -1,5 +1,6 @@
 import { useEffect, useState, useCallback, useRef } from "react";
 import { useRouter } from "next/router";
+import { usePathname, useSearchParams } from "next/navigation";
```
sital002, line 3: we are using nextjs pages router ```next/navigation``` is for app router. Please use router from ```next/router```

badrivlog: Thanks 👍 remove `next/navigation` and made necessary changes

pages/search.js
```diff
@@ -252,12 +249,12 @@ export default function Search({
               ref={searchInputRef}
               placeholder="Search user by name or tags; eg: open source, reactjs or places; eg: London, New York"
               name="keyword"
-              defaultValue={userSearchParam}
+              defaultValue={searchTerm}
             />
           </form>
```
sital002, line 254:
```suggestion
          <Button className={"my-2"}>Search</Button>
          </form>
```

sital002:
Test on local environment works well 
can you add the missing dependency on the useEffect https://github.com/EddieHubCommunity/BioDrop/pull/9248/files/0db01c52abb61e2ac56b8a1fa776d825fbb8e50f..0272140c79a7bfe45eb269408e7fc2138d34faca#diff-99c55e6229f7a0d904e59da5da6e77729e62b05852484cf2afff2e13292b50b1R147

pages/search.js
badrivlog: Screenshot (49)

added the search button. not designed a good-looking search bar because this PR is more focused on search functionality. search bar re-design could come in another PR as I don't want to add more complexity to this PR, Thanks

eddiejaoude (approved):
Looks great, thank you 👍

PR #9262, opened 2023-10-02
fix: disable the reorder button unless there is more than one link added

### LINKED ISSUE #9221: [BUG] Reorder button should be visible only if multiple links added
### Description

Edit:
~~We can hide~~ `Disable` the reorder button unless there is more than one link added.

### Screenshots

reorder

Yes

### Additional information

_No response_

sanketmp:
## Fixes Issue
Fixes #9221 

## Changes proposed
- Added condition to disable/enable the reorder button depending on number of links added.

## Screenshots
https://github.com/EddieHubCommunity/BioDrop/assets/68698332/5b7eff77-7c92-4b2c-a24a-800559e962ff

## Note to reviewers
_(attached video instead of screenshot)_

### DIFF
--- a/pages/account/manage/links.js
+++ b/pages/account/manage/links.js
@@ -111,7 +111,7 @@ export default function ManageLinks({ BASE_URL, username, links }) {
             Add Link
           </Button>
 
-          {linkList.length !== 0 && !reorder && (
+          {!reorder && (
             <Button onClick={() => setReorder(true)} disable={linkList.length < 2}>
               <ArrowPathIcon className="h-5 w-5 mr-2" />
               REORDER
--- a/pages/account/manage/links.js
+++ b/pages/account/manage/links.js
@@ -112,7 +112,7 @@ export default function ManageLinks({ BASE_URL, username, links }) {
           </Button>
 
           {linkList.length !== 0 && !reorder && (
-            <Button onClick={() => setReorder(true)}>
+            <Button onClick={() => setReorder(true)} disable={linkList.length < 2}>
               <ArrowPathIcon className="h-5 w-5 mr-2" />
               REORDER
             </Button>


pages/account/manage/links.js
```diff
@@ -112,7 +112,7 @@ export default function ManageLinks({ BASE_URL, username, links }) {
           </Button>
 
           {linkList.length !== 0 && !reorder && (
-            <Button onClick={() => setReorder(true)}>
+            <Button onClick={() => setReorder(true)} disable={linkList.length < 2}>
```
hi-Kartik2004, line 115: Maybe you could add `disabled` to the reorder button on the disable condition, so that it's easier to figure out that the reorder button is disabled, along with making the pointer events to be none.

Refer the below syntax for details

```
     <button type="button" className="some tailwind classes" disabled>Button</button>
```

### This is how the disabled button will look after the change (ofcourse, pointer events will still remain none)
image

sanketmp: Button is a custom component defined in `components/Button.js` which is imported in `manage/links.js` and `disable` is a prop not html attribute.
And using `disabled` directly on `<button> in components/Button.js` will not check condition.

eddiejaoude, line 115: From the line above, I think we need to remove this so it is displayed? `linkList.length !== 0`

sanketmp: `linkList.length !== 0` is for displaying/hiding reorder button. if we removed that, it will display _disabled_ reorder button even if there are no links. 

`linkList.length !== 0`will check if any links are there  and  `disable={linkList.length < 2}`will check how many and then disable/enable. So we need both.

eddiejaoude: Yes you discussed this with our maintainer here https://github.com/EddieHubCommunity/BioDrop/issues/9221#issuecomment-1742009718

sanketmp: okay and will do the changes.

eddiejaoude (approved):
Thank you 👍

PR #9513, opened 2023-10-16
Fixed : Docs URL refresh when using side nav issue #9486

### LINKED ISSUE #9486: [BUG] Docs URL refresh when using side nav
### Has this bug been raised before?

### Where did you find this bug?

production on biodrop.io

### Version of BioDrop (for example "v1.2.3")

Current version v2.72.9

### Description

Most of the pages on biodrop.io load without refreshing the URL, but in the documentation module if I click any of the links from the left sidebar it refreshes the URL. ()

I can fix this, all documentation pages will load without refreshing the URL.

### Screenshots

Image of the links:
https://github.com/EddieHubCommunity/BioDrop/assets/49125977/5e787d09-b259-47c3-bb00-60130763c446

Video of the issue:
Screenshot_1

Yes

### If you want to work on this issue...

Issue in code:
before-issue

How to solve this : 
issue_solve

I added the "next/link" to solve the URL refresh problem, And it works perfectly.

devmamunur:
## Fixes Issue
Closes #9486 

## Changes proposed

Now, the docs side-nav link click will not refresh the full page. All docs pages will load without refreshing the URL.

## Video

https://github.com/EddieHubCommunity/BioDrop/assets/49125977/8ef33efd-52c7-475e-b14b-667c4139bab7

### DIFF
--- a/components/navbar/SideNav.js
+++ b/components/navbar/SideNav.js
@@ -62,8 +62,7 @@ export default function SideNav({ navigation }) {
                         {item.children.map((subItem) => (
                           <li key={subItem.name}>
                             {/* 44px */}
-                            <Disclosure.Button
-                              as="a"
+                            <Link
                               href={subItem.href}
                               className={classNames(
                                 subItem.href.toLowerCase() == pathname &&
@@ -72,7 +71,7 @@ export default function SideNav({ navigation }) {
                               )}
                             >
                               {subItem.name}
-                            </Disclosure.Button>
+                            </Link>
                           </li>
                         ))}
                       </Disclosure.Panel>


amandamartin-dev (approved):
Tested in Gitpod works as expected.  This fix is also what is recommended for working with Next from headless UI github issue convo.  Looks good to me, thanks for the contribution

components/navbar/SideNav.js
```diff
@@ -62,8 +62,7 @@ export default function SideNav({ navigation }) {
                         {item.children.map((subItem) => (
                           <li key={subItem.name}>
                             {/* 44px */}
-                            <Disclosure.Button
-                              as="a"
+                            <Link
```
eddiejaoude, line 65: I think it would be better to use `as` they have recommended in the docs 
https://headlessui.com/react/disclosure#rendering-a-different-element-for-a-component

eddiejaoude (changes requested):
I saw from the docs they recommend using `as`, interested to know people's thoughts?

amandamartin-dev:
> I saw from the docs they recommend using `as`, interested to know people's thoughts?

Yes, using "as" recommended and what we had previously, but if you look further this issue is common when using the next framework and in such case is recommended to use a Link component

In reading through the Github Issue here [Refresh Issue](https://github.com/tailwindlabs/headlessui/issues/453)

is how I came to the conclusion that this fix is needed and appropriate in this case as it seems to be an issue within the Headless UI components.

wdyt?  perhaps I missed something?

components/navbar/SideNav.js
devmamunur: When using the 'as="a"' attribute, it results in a URL refresh. Thus far, I have not found a solution to this issue when using this approach. The common advice I've come across suggests using 'Link' instead of 'as="a"'. It seems that 'as="a"' does not work optimally in this specific scenario

devmamunur:
Hi, @eddiejaoude  and @amandamartin-dev 

I found a new solution to fix this issue. we don't need to use the next/link, we can use headless UI `as`. For that we need to do a small hack.

devmamunur:
Hi, @eddiejaoude and @amandamartin-dev

I found a new solution to fix this issue. we don't need to use the next/link, we can use headless UI as. For that we need to do a small hack.

devmamunur:
Hi, @eddiejaoude and @amandamartin-dev

I found a new solution to fix this issue.  We can use headless UI `as` For that we need to do a small hack.  we can prevent this issue using a custom function when clicking the link we need to use preventDefault(), then push the link. 
step one           
step-2

eddiejaoude:
>  small hack

Thank you for looking at alternative solutions, but I don't like the idea of a "small hack" 🤔

amandamartin-dev:
> > I saw from the docs they recommend using `as`, interested to know people's thoughts?
> 
> Yes, using "as" recommended and what we had previously, but if you look further this issue is common when using the next framework and in such case is recommended to use a Link component
> 
> In reading through the Github Issue here [Refresh Issue](https://github.com/tailwindlabs/headlessui/issues/453)
> 
> is how I came to the conclusion that this fix is needed and appropriate in this case as it seems to be an issue within the Headless UI components.
> 
> wdyt? perhaps I missed something?

@eddiejaoude bumping my original comment about the initial version of this in the PR since this has been open for a bit now (not the update with the "small hack").  Take a look and lmk what you think

eddiejaoude:
I like your idea Amanda for using the Link component, thank you for reminding me 👍

eddiejaoude (approved):
Great collaboration 💪

PR #9652, opened 2023-10-28
Fix [BUG] Icon copy notification text is not readable as the underlying elements are also visible.

### LINKED ISSUE #9618: [BUG] Icon copy notification text is not readable as the underlying elements are also visible.
### Has this bug been raised before?

### Where did you find this bug?

production on biodrop.io

### Version of BioDrop (for example "v1.2.3")

latest version

### Description

In the icons page when we click on the icons then that icons gets copied and we see a toast notification on the right side of the screen. However, the toast notification alert is not solid and hence the underlying elements are not showing visible through the notification and hence the text written in the notification is often not readable properly.

I am attaching a screenshot below.

I think if we can have a solid toast notification that will be better in my opinion. 

### Screenshots

image

No

### If you want to work on this issue...

_No response_

mrinmay7875:
Closes #9618 

## Fixes Issue
This PR fixes #9618 

## Changes proposed

- Removed the unnecessary z index from navbar and now the Icon copy notification shows over the navbar as it should and text in the icon notification is readable properly.

## Screenshots
Here is a screenshot after my fix:
image

### DIFF
--- a/components/navbar/Navbar.js
+++ b/components/navbar/Navbar.js
@@ -135,11 +135,11 @@ export default function Navbar() {
         className={classNames(
           "relative top-0 bg-primary-high dark:bg-primary-medium",
           session &&
-            session.accountType === "premium" &&
-            "border-b-2 border-tertiary-medium",
+          session.accountType === "premium" &&
+          "border-b-2 border-tertiary-medium",
         )}
       >
-        <div className="z-30 w-full mx-auto px-4 sm:px-6 lg:px-8 relative t-0">
+        <div className="w-full mx-auto px-4 sm:px-6 lg:px-8 relative t-0">
           <div className="flex items-center justify-between h-16">
             <div className="flex items-center">
               <div className="flex-shrink-0">
@@ -229,7 +229,7 @@ export default function Navbar() {
             isOpen
               ? "transform translate-y-0 opacity-100"
               : "transform -translate-y-96 opacity-0",
-            "md:hidden z-20 absolute t-0 bg-primary-medium transition-all duration-700 ease-in-out w-full",
+            "md:hidden absolute t-0 bg-primary-medium transition-all duration-700 ease-in-out w-full",
           )}
           id="mobile-menu"
         >


Kamaruddheen:
Hey @mrinmay7875, I was thinking, what if we bumped up the z-index for the notification or alert instead of removing it from the navbar? Removing the navbar's z-index might potentially affect other components, and I thought this could be a more targeted solution. What do you reckon?

mrinmay7875:
@Kamaruddheen I did test it well but could not find any area where it is broken due to removing the z-index from navbar. Let's see what the maintainers say about this.

Kamaruddheen:
It should be possible if we are generating a alert template from our end.

Kamaruddheen:
I could see alert.js file which is generating alerts based on success, errors, warning and info. If we increase the z-index of those class we should also get the same result.

Kamaruddheen:
> This issue relates to the Notification component and its interaction with the IconCard component, not the Alert component!

After some investigation, I've pinpointed the actual bug. It turns out that z-40 is already defined in the Notification component, specifically on line no. 54

https://github.com/EddieHubCommunity/BioDrop/blob/64fa44d85048d04fd7f097e2f6f8a1f5c31f7144/components/Notification.js#L52-L72

I attempted to include the z-index within a `<div>` element of the Transition component, but it seems that the **z-index is having an impact due to the presence of position: static.** I looked further, I couldn't identify any instances of `position: static` within the `<div>` elements.

mrinmay7875:
I am busy with some personal stuff. @Kamaruddheen You can take this over if you want and create a PR with your changes.

Kamaruddheen:
Okay @mrinmay7875, I think I might be able to commit in this PR. Let me try to commit to this PR branch.

SaraJaoude:
@mrinmay7875 if you cannot work on an issue please unassign yourself. I will do this for you on this occasion. 

@Kamaruddheen committing to someone else's PR would cause confusion and extra admin, so please raise a new PR. I will assign the issue to you. 

As a result of the above conversation I am closing this PR.

PR #9689, opened 2023-10-30
data: leetebbs

leetebbs:
This is my profile creation with json.


sital002 (commented):
Looks like you already have created a biodrop profile using forms there is no point in creating the profile for json Here is the link for your biodrop profile using forms https://www.biodrop.io/leetebbs .
I will be closing this PR.

PR #9878, opened 2023-12-04
feat: group suggestions for users link 

### LINKED ISSUE #9836: [FEATURE] Users should be able to select from already existing groups
### Is this a unique feature?

### Is your feature request related to a problem/unavailable functionality? Please describe.

Unavailable functionality: If a user already typed in a group name for links, the user shouldn't need to type the same name again when assigning links to that group.

### Proposed Solution

I think after a group name is been typed in at first, next time the user is adding a link there should be options for user to either choose from existing group names or type a new one.

### Screenshots

_No response_

No

### If "yes" to above, please explain how you would technically implement this

Would have love to implement this but not my tech stack, and I don't want to try and error
Here is what I propose though

Using the autocomplete library in JavaScript
1 - Make an empty array
2 - Save the names of group provided by the user if they don't already exist
3 - Put this names in the array and pass it to the autocomplete library to show them when ever users start typing

sital002:
## Fixes Issue

closes #9836 

## Changes proposed

Users will be able to view all the previous groups.

## Screenshots
After changes:
image
image


pages/api/account/manage/link/[[...data]].js
```diff
@@ -55,6 +54,24 @@ export async function getLinkApi(username, id) {
...
+  let getGroupLink = [];
+  getGroupLink = await Link.aggregate([
+    { $match: { username } },
+    { $group: { _id: null, groups: { $addToSet: "$group" } } },
+    { $project: { _id: 0, groups: 1 } },
+  ]).exec();
+  getGroupLink = getGroupLink[0].groups.filter((item) => item !== "");
+
+  if (!getGroupLink || getGroupLink.length === 0) {
+    log.info(`Grouo not found for username: ${username}`);
```
Comment (eddiejaoude) [inline pages/api/account/manage/link/[[...data]].js:69]:
```suggestion
    log.info(`Group not found for username: ${username}`);
```

components/GroupLinkSearch.js
```diff
@@ -0,0 +1,52 @@
...
+          .slice(0, 5);
+
+  return (
+    <Combobox value={selectedGroup} onChange={handleGroupSelection}>
+      <Label htmlFor="search-groups">Group</Label>
+      <Combobox.Input
+        name="search-group"
+        onChange={handleGroupChange}
+        className={`border-2 transition-all duration-250 ease-linear rounded px-6 py-2 mb-2 block w-full dark:bg-primary-high hover:border-tertiary-medium focus:ring-0 focus:border-tertiary-medium focus:outline-0`}
+      />
```
badrivlog, line 32: It would be better if we associated the `label` with the `input` field explicitly by using `id` attribute. This enhances accessibility by establishing a clear connection between `label` and `input`.

The following resources could be helpful: https://www.w3.org/WAI/tutorials/forms/labels/#associating-labels-explicitly

components/GroupLinkSearch.js
```diff
@@ -0,0 +1,52 @@
...
+          })
+          .slice(0, 5);
+
+  return (
+    <Combobox value={selectedGroup} onChange={handleGroupSelection}>
+      <Label htmlFor="search-groups">Group</Label>
+      <Combobox.Input
+        name="search-group"
+        onChange={handleGroupChange}
+        className={`border-2 transition-all duration-250 ease-linear rounded px-6 py-2 mb-2 block w-full dark:bg-primary-high hover:border-tertiary-medium focus:ring-0 focus:border-tertiary-medium focus:outline-0`}
```
badrivlog, line 31: not sure why template literal? i think we could use plain string

sital002: I copied this from IconSearch component since they are both similar.

badrivlog: In our project, we've typically used template literals when dealing with dynamic values. Since the value here seems static, a plain string might enhance readability and maintain consistency with our coding style

eddiejaoude: Great collaboration 💪

eddiejaoude (approved):
Works well, thank you 👍

components/GroupLinkSearch.js
eddiejaoude: If this still needs improving, could you raise as a separate issue, thank you 👍

badrivlog: Sure, I'll create a new issue to address the improvement. Thanks for your guidance! 👍

PR #9912, opened 2023-12-15
Add the possibility to reorder the repositories

### LINKED ISSUE #9810: [FEATURE] Re-ordering repos
### Is this a unique feature?

### Is your feature request related to a problem/unavailable functionality? Please describe.

When I want to change the order in which my repos are displayed I have to delete and insert them in the order I want, it's a bit long of a process and one easily give up on that (or at least that's my case)

### Proposed Solution

That would be nice a re-ordering function, just like the one we use for re-ordering the links!
It's beautiful and usable and would allow to better showcase our most relevant repos on the top 

### Screenshots

_No response_

Yes

### If "yes" to above, please explain how you would technically implement this

I would implement the <ReactSortable> wrapper just like in the links page, that could be enabled or disabled, so we could change the links order. That would also trigger a save method to save the new order.

Giuliano1993:
## Fixes Issue

Closes #9810

## Changes proposed

I added the possibility to reorder the user repository.
When the repos are returned, they are retrieved in saved order.
The mechanic used is the same as for user links reordering, just adjusted it a bit

## Screenshots

Screenshot 2023-12-15 222102
Screenshot 2023-12-15 222122
Screenshot 2023-12-15 222147


components/user/UserRepos.js
```diff
@@ -81,10 +115,47 @@ export default function UserRepos({ manage = false, confirmDelete, repos }) {
...
     <ul role="list" className="divide-y divide-primary-low">
-      {repos.map((repo) => (
+      <ReactSortable
+          list={reposList}
+          setList={setReposList}
+          disabled={!reorder}
+          ghostClass="border-2"
+          chosenClass="border-dashed"
+          dragClass="border-red-500"
+        >
```
badrivlog, line 153: some tests failed because `<li>` elements must be contained in a `<ul>` or `<ol>`, but here `<li>` contained by `<div>` element.
to fix this please follow the `react-sortablejs` docs: https://www.npmjs.com/package/react-sortablejs#tag

Screenshot (56)

badrivlog (commented):
Please install vscode prettier extension to remove extra blank line

components/user/UserRepos.js
Giuliano1993: Oh! thank you for the review! It should be okay by now! 
Let me know if any further adjustment is necessary

pages/api/account/manage/repos.js
```diff
@@ -8,15 +8,52 @@ import Profile from "@models/Profile";
...
     return res.status(400).json({ error: "Invalid request: GET required" });
   }
 
-  const data = await getReposApi(session.username);
+  const username = session.username;
+  let data = [];
+  if(req.method === "GET"){
+    data = await getReposApi(username);
+  }
+  //const data = await getReposApi(username);
```
eddiejaoude, line 20: Can this commented out line be removed?

pages/api/account/manage/repos.js
```diff
@@ -8,15 +8,52 @@ import Profile from "@models/Profile";
...
+        },
+      );
+    } catch (e) {
+      log.error(e, `failed to update repo order for username: ${username}`);
+    }
+  });
+
+  const repos = await Promise.allSettled(repoList).then(() => { 
+    return getReposApi(username);
+  })
```
eddiejaoude, line 52: Is this needed? Because the map and content are already using `await`

eddiejaoude (changes requested):
Looks and works great! 👍  Just some small inline comments

I notice when adding a repo, it doesn't appear on the list until the page is refresh, but this bug doesn't exist in production

pages/api/account/manage/repos.js
Giuliano1993: Actually this yes is needed because, if we don't wait for all the promises to be settled, the reorder happens, but when you save the returned order is the same it had before, because it hasn't finished saving yet. If you refresh the page you actually see the correct order though. 
In this way it shows you the right order as you save.
Don't know if I'm missing some node points her  or maybe on server works, but locally it doesn't do the trick without specifying to wait all the ordering promises to be settled

Giuliano1993:
Hi @eddiejaoude checked out the adding part and that is true! Working to fix that!

pages/api/account/manage/repos.js
eddiejaoude: Ok, so I think if we need the promise all, then the map doesn't need the `await`? 

https://github.com/EddieHubCommunity/BioDrop/pull/9912/files/6082f0485096f7d9a8c4a00d422c4bc1058666ad#diff-0cf4a906bb37e0a3ff6060fb19d418de9313848bbbcf750c0effc667750bec9dR32

```diff
- const repoList = await data.map(async (repo, idx) => {
+ const repoList = data.map(async (repo, idx) => {
```

and possibly here also?

```
return await Profile.findOneAndUpdate(
```

Giuliano1993: Yes indeed those weren't needed after all, removed them and it is still working! Thank you for pointing that out!

Giuliano1993:
Okay @eddiejaoude i fixed all previous issues, it should be okay by now!

eddiejaoude:
Sorry, on this now 👍

pages/account/manage/repos.js
```diff
@@ -72,7 +71,7 @@ export default function ManageRepos({ BASE_URL, repos }) {
 
     const resRepos = await fetch(`${BASE_URL}/api/account/manage/repos`);
     const listRepos = await resRepos.json();
-
+    console.log(listRepos);
```
eddiejaoude, line 74: I think this can be removed?

```suggestion
```

eddiejaoude (changes requested):
Thank you for making the changes. I just tested it and on the profile page the repos are not ordered, I think this API also needs a change `pages/api/profiles/[username]/index.js`?

Giuliano1993:
Sorry for the delay! Yes you're right, fixing it rn!

eddiejaoude (approved):
Works great 👍  Thank you

PR #10127, opened 2024-01-11
Fix 9878 group suggestion link 

sital002:
## Fixes Issue

fixes  #9878 


sital002:
@eddiejaoude  the https://github.com/EddieHubCommunity/BioDrop/pull/9878 pr is breaking the links page i reverted the changes and made the fix can u look into this

sital002:
The previous PR would break the add new link page if i already have links but don't have any group since i am returing a object if no group found [here](https://github.com/EddieHubCommunity/BioDrop/pull/9878/files#diff-fca4d78bd846ab4a603e34aa801b1468713aaad07f3b15beb07a099398f62d44R70)  and trying to call slice method on the object [here] (https://github.com/EddieHubCommunity/BioDrop/pull/9878/files#diff-4e4ef483debf9877cce8671c40896f1ed16ae09562ecad73c37aa143a1e23257R18-R23)  i should have returned an array instead. 
I have fixed those changes here.

pages/api/account/manage/link/[[...data]].js
```diff
@@ -55,6 +54,18 @@ export async function getLinkApi(username, id) {
...
 
+export async function getGroupLinkApi(username) {
+  await connectMongo();
+  let getGroupLink = [];
+  getGroupLink = await Link.aggregate([
+    { $match: { username } },
+    { $group: { _id: null, groups: { $addToSet: "$group" } } },
+    { $project: { _id: 0, groups: 1 } },
+  ]).exec();
+  getGroupLink = getGroupLink[0].groups.filter((item) => item !== "");
```
Comment (eddiejaoude) [inline pages/api/account/manage/link/[[...data]].js:65]: Does this need a check for `.length`?

components/GroupLinkSearch.js
```diff
@@ -0,0 +1,60 @@
+import { useState } from "react";
+import { Combobox } from "@headlessui/react";
+import Label from "./form/Label";
+
+function GroupLinkSearch({ selectedGroup, handleGroupSelection, groups }) {
+  const [query, setQuery] = useState("");
+
+  const handleGroupChange = (e) => {
```
ThomasCode92, line 8: I find _e_ a bit ambiguous, consider renaming it to `event`. This can enhance code readability and understanding.
Renaming it to `event` makes it clear that this an _event handler_.

pages/account/manage/link/[[...data]].js
```diff
@@ -30,6 +34,12 @@ export async function getServerSideProps(context) {
   const id = context.query.data ? context.query.data[0] : undefined;
 
   let link = {};
+  let groups = [];
```
Comment (ThomasCode92) [inline pages/account/manage/link/[[...data]].js:37]: Maybe `linkGroups` is a beter name, than it's clear what kind of groups this variable holds

pages/api/account/manage/link/[[...data]].js
Comment (sital002) [reply pages/api/account/manage/link/[[...data]].js:65]: I don't think this need a check for length the getGroupLink returns a a array with group as it's object
```js
getGroupLink = [ { groups : [""] } ];
```
even if the query don't find any goup this will still return the data in the format i mentioned above.

Comment (sital002) [reply pages/api/account/manage/link/[[...data]].js:65]: You can verify this by logging the values.

Comment (eddiejaoude) [reply pages/api/account/manage/link/[[...data]].js:65]: ok thank you 👍

components/GroupLinkSearch.js
eddiejaoude: I think we use `e` in other places, so should not be an issue, but thank you for highlighting

eddiejaoude (approved):
Thank you

Issue #10141, opened 2024-01-13
[FEATURE] Adding onClick prop in the Button component for better intellisense


nidhish-srivastava:
### Is this a unique feature?

- [X] I have checked "open" AND "closed" issues and this is not a duplicate

### Is your feature request related to a problem/unavailable functionality? Please describe.

The button uses onClick event very often and in the Button component,we dont have a prop for onClick

### Proposed Solution

In the Button component,we have ...restProps but i will add an onClick prop for better developer experience and intellisense

### Do you want to work on this issue?

Yes

SaraJaoude:
Thank you for raising.  I have assigned this to you. 

Please also add `href` as that is used just as frequently.

PR #10156, opened 2024-01-20
feat : Button component onClick and href prop

nidhish-srivastava:
## Fixes Issue 
#10141

## Changes proposed

Added onClick and href  prop to the button component since they are the most used props while using the Button component so we get better intellisense while developing.

### DIFF
--- a/components/Button.js
+++ b/components/Button.js
@@ -6,6 +6,8 @@ export default function Button({
   disabled = false,
   className,
   overrideClassNames = false,
+  href,
+  onClick,
   children,
   ...restProps
 }) {
@@ -34,6 +36,8 @@ export default function Button({
 
   const button = (
     <button
+    href={href}
+    onClick={onClick}
       className={
         overrideClassNames ? className : classNames(defaultClassName, className)
       }


components/Button.js
```diff
@@ -34,6 +36,8 @@ export default function Button({
 
   const button = (
     <button
+    href={href}
```
ArvindParekh, line 39:
```suggestion
```

Shouldn't we move the href prop to `Link`?
It seems like the intention was to use the `href` prop to determine whether to render a `Link` or a `button`. However, the return statement (`return restProps.href ? link : button;`) will always return a `link` (without an `href`) in this case.

Maybe we should move the `href` prop to the `Link` component. This way, when `href` is present, the `Link` component will be rendered with the `href` attribute, and when `href` is not present, a `button` will be rendered instead.

```
<Link
      href={href}
```
Also we should change the return statement to `href ? link : button;`

ArvindParekh:
Also, could you please edit your PR description and link your issue below the heading, with a hash, like this: `#10141`

components/Button.js
nidhish-srivastava: i was thinking of the same thing as well

PR #10165, opened 2024-01-22
refactor: button component

sital002:
## Changes proposed
Refactor button component
1. Added cva and tailwind merge 
2. Added jsdocs for type safety 
3. currently we can't use ref on the button component so using forward ref so that the button component can be selected.

### DIFF
--- a/components/Button.js
+++ b/components/Button.js
@@ -1,48 +1,65 @@
+import React from "react";
 import { classNames } from "@services/utils/classNames";
 import Link from "./Link";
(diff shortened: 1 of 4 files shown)


sital002:
Currently the all the button have default style if you approve that we should implement these library and use js doc for type safety I will replace all the button props from primary=true to variant=primary 
What do you think @eddiejaoude If we are planning to use Typescript in the future then we don't need js doc but if not i think we should use js doc where it makes more sense

